PR #2, opened 2016-12-13
Fix contributing docs heading

hflynn:
Fix duplicated heading under Support


hflynn:
Merging as I'm starting work on these pages again today.

PR #7, opened 2017-04-07
OMERO Product pages

hflynn:
See https://trello.com/c/E7FOdYBQ/36-omero-product-pages, https://trello.com/c/rNxyByyQ/34-re-write-users-overview-tab, and https://trello.com/c/bcKP9wuv/33-omero-tools for context

This PR builds on Liza's work to move the features to the same level of the scientists/dev/admins split - adding additional feature group pages, adding community tools to those pages as well as core OMERO features and linking the feature groupings from the user-type summary pages.


hflynn:
staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/products/omero/

pwalczysko:
From the Downloads page
screen shot 2017-04-10 at 10 37 15
The long blue text needs to be split by a black text, as there are two distinct clickable links under that.

pwalczysko:
As discussed with @hflynn already, the image on the front page is so huge that the user does not realize the whole website is not just the "Downloads" button in the middle of it. On any other screen than the huge Thunderbolt, the features like 
screen shot 2017-04-10 at 10 40 01

are not easily read, as the user needs to understand that some scrolling is needed.

pwalczysko:
For Big images section the link http://help.openmicroscopy.org/viewing-data.html#viewer seems to me better than http://help.openmicroscopy.org/viewing-data.html. As there is no mention at all in Help about Big images, but still, it makes much more sense to navigate to the subchapter of Viewing images which deals with Full Viewer (you cannot pan and zoom (mentioned in the text) too much in Preview, and cannot see big images in Preview of insight at all).

screen shot 2017-04-10 at 10 53 14

pwalczysko:
screen shot 2017-04-10 at 11 08 28
This needs some reformulation. Not that it is not true, it is just very hard to read it right and without confusion. Maybe:
``Group owners and sysadmins can import data for other users. The data will be owned by the users they imported for.``

pwalczysko:
screen shot 2017-04-10 at 11 24 04
A bit too little content in this description. I guess the biggest strenght of smuggler is that it can import "offline", ie. after the user left the facility, but maybe @jburel has a better idea.

pwalczysko:
screen shot 2017-04-10 at 11 27 44
We were asked by users to rename the feature so that the confusion with the DropBox software is prevented. The label unfortunately just add to this confusion.

pwalczysko:
Also analysis results are being saved from IJ/Fiji to OMERO - I think this should be in the blurb.
screen shot 2017-04-10 at 11 36 28

pwalczysko:
screen shot 2017-04-10 at 11 39 11
I would not mix Export and ImageJ. For example "Export to ImageJ" sounds strange, and is not accurate in the sense that no menu called Export to ImageJ (or to ImageJ format) is available.

hflynn:
@pwalczysko I think the downloads page is still a work in progress so I'm not going to touch it on this PR, likewise the image banner size.

I'll push a commit for the features issues though, thanks!

jburel:
smuggler: key features: ability to queue import for offline import

jburel:
smuggler: key features: ability to queue import for offline import

pwalczysko:
screen shot 2017-04-10 at 14 20 45

Link first to Balaji's videos, such as http://downloads.openmicroscopy.org/movies/omero-5-2/mov/fiji_scripting_demo.mov

pwalczysko:
Missing OMERO.matlab binding in this section below.
screen shot 2017-04-10 at 14 23 38

Also point to the video http://downloads.openmicroscopy.org/movies/omero-5-2/mov/OMERO_Matlab.mov

pwalczysko:
screen shot 2017-04-10 at 14 26 32
This section for me should be quite high above actually, just the second one after the insight import I would think.
Also, link the two videos of Balaji to it, http://downloads.openmicroscopy.org/movies/omero-5-2/mov/OMERO_CLI_import_basic.mp4 and http://downloads.openmicroscopy.org/movies/omero-5-2/mov/OMERO_CLI_import_advanced.mp4

pwalczysko:
screen shot 2017-04-10 at 14 35 06
I would not use "bespoke", neither me nor Balaji nor apparently many non-British native speakers know what that means - I suppose "customized" or something of that sort ?

pwalczysko:
screen shot 2017-04-10 at 14 39 02
Move this to "Publish" section. 
Add to this Export section a Download - (meaning Download of original data), where we should also highlight that this will download whole MIFs (or not if you use Web, you can pick and mix in Web) and that you can download whole plates (configurable on server, not by default, but functioning). Of course, I do not suggest all this info must be in the front page blurb, but I suppose I should be albe to click myself through to it.

pwalczysko:
screen shot 2017-04-10 at 14 45 04
Would put this as the 3rd blurb on the page (just after the "modify groups and users in web" and "Use CLI for admin tasks). Out of experience this is very important feature and widely used.

pwalczysko:
Couple of edits, otherwise I like it a lot, very sleek and powerful.

hflynn:
@pwalczysko @jburel that's another round of edits pushed and https://snoopycrimecop.github.io/www.openmicroscopy.org/products/omero/ updated.

As well as Petr's comments, I moved mtools into the main analysis section and added R in since we have both a movie and a repo to link to.

pwalczysko:
The changes are good, all reads very nicely for me now. Ready to merge FMPOV.

jburel:
the only change I will make is "R and OMERO" instead of rOMERO

hflynn:
@jburel pushed that change

PR #8, opened 2017-04-12
Dev nav

lunson:
Created a separate navigation menu for Developers section so more content can be added without affecting the menu for the Scientists and Institution.


hflynn:
👍 thanks

hflynn:
This works fine built locally and I see how to add extra pages now so merging

PR #10, opened 2017-04-28
Fix user group pages

lunson:
Fixes the background img issue for the user group subpages. This is based off of the https://github.com/lunson/www.openmicroscopy.org/tree/add-visuals branch.


hflynn:
You could push these changes to the add visuals branch @lunson - you don't need to open another PR when the first one isn't merged

hflynn:
this does fix the issue though 👍

hflynn:
Actually, since all the commit are here, might be easier if I just close the first PR and we rename this one

hflynn:
You could cherry pick the last commit to your origin branch and then push the fix there rather than us having two PRs open with the same changes on

PR #11, opened 2017-04-28
Explore page

hflynn:
See https://trello.com/c/V0b1FtwP/18-getting-started

This is a first stab at an 'Explore OME in Action' page

Still to do:

- [ ] Add logos/images
- [ ] Improve formatting
- [ ] Add embedded image viewer
- [ ] Make decision about whether to feature a developer section (I suggest we put developer content under the omero/develop section instead)

Figure demo will also need migrating at some point but can be done with the rest of the content we want to save from figure.openmicroscopy.org (see https://trello.com/c/U9PIYgHx/40-figure-openmicroscopy-org-site-migration)


hflynn:
Staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/explore/

explore/index.html
```diff
@@ -0,0 +1,194 @@
...
+        <!-- begin section for OMERO for Scientists -->
+        <!-- Liza to add OMERO, Imagej2, IDR and an OMERO.figure logo/same graphics as used to the figure feature, and reformat as needed -->
+        <hr class="whitespace">
+        <div class="row">
+            <div class="small-12 columns">
+                <h2>OME for Scientists</h2>
+                <h5 class="subheader">Whether you have your own microscopy
+                    images or want to test functionality with
+                    publicly available data, there's a tool or demo for you to
+                    try</h5>
```
mtbc: end with period?

will-moore:
I think this page is better with the rename from "Getting Started", but I'm still not exactly clear what the purpose of this page is? I think that it has some useful content but I think users could spend a lot of time on the site before they happen to click on "Explore". It's also a mixture of "Getting started" links and "live demos".
If a user comes to the site wanting to know more about OMERO or BioFormats they will likely follow the links from the home page to the OMERO/BF sections. 

On the current https://www.openmicroscopy.org/site/ there's a "New Users" link from the home page, and this is really a single "Getting Stared" page for users/sysadmins etc. We don't really have an equivalent of that in the new site. 
Maybe we need an "Explore OMERO" and "Getting Started" under the OMERO section of the home page. 

Under the Bio-Formats section it would really just be "Getting Started" since there are no "Explore" links for BF. The current link under BF on the home page is "Get started with Bio-Formats →" but the page it links to doesn't have any "Getting started" info, so it would benefit from the link to "Bio-Formats ImageJ plugin".

If we had an "Explore OMERO" page (without the BF and Demo-server "getting started" links), then I think there are 3 categories of links:

 - Demo of webclient OMERO.web: IDR, QBI, SSBD
 - Demo of other apps: OMERO.figure, OMERO.gallery (https://cci02.liv.ac.uk/gallery/)
 - Public repos with custom front-ends


Minor feedback on the current page:

 - It seems that everything is left-justified whereas most pages in the site are centred.
 - Re "Hosting images doesn't have to stop at sharing links to OMERO.web though. With our range of apps for OMERO.web, you can also embed image viewers in your institutional websites" Since this is a demo page, we should show an example of this. What about embedding an example in the page itself?
Jason's GRE page doesn't have an embedded example from nightshade anymore.
Also the double-negative "doesn't have to stop" is less promotional than a positive. e.g. "You can also..."

jburel:
one example of gallery: https://cci02.liv.ac.uk/gallery/

jburel:
with the first version, let's try to be minimal:

 * link to demo account
 * collection of external links: the ones you have + gallery at liverpool. So we can show that omero (or part of it) in used
I will not distinguish between app and full omero.

hflynn:
@jburel https://snoopycrimecop.github.io/www.openmicroscopy.org/explore/ better for now or did you want the 'OME for Scientists' and 'OME at Institutions' headings removed too for now?

jburel:
I will keep the sections as they are and we can add more content later on

hflynn:
@lunson what is the easiest way to move forward with this for you? Shall I merge it and leave you to tidy up the formatting and add images when you get time or would you like me to show you how to work on top of my branch?

lunson:
@hflynn Merge, please. Thank you!

PR #12, opened 2017-04-28
Add logos and bg imgs

lunson:
Adding logos and background images. I will redo OME-Files if needed after a decision has been made.


hflynn:
@lunson from @sbesson comment on https://trello.com/c/zJMmuwyl/42-ome-files-vs-ome-files I think we can safely say the hyphen can go. Many apologies if my old trello card was the original source of confusion!

hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org updated

PR #15, opened 2017-05-09
Bio-Formats 5.5.0

sbesson:
As discussed with @jburel @hflynn and @lunson, this PR updates Bio-Formats pages to match the latest Bio-Formats 5.5.0

### DIFF
--- a/products/bio-formats/downloads/index.html
+++ b/products/bio-formats/downloads/index.html
@@ -74,7 +74,7 @@ title: Bio-Formats Downloads
                                             <h5>API Documentation</h5>
                                             <h6>bio-formats-javadocs-5.5.0.zip</h6>
                                             <h6>Checksum: ced3ffb7 (<a href="https://downloads.openmicroscopy.org/bio-formats/5.5.0/artifacts/bio-formats-javadocs-5.5.0.zip.sha1">SHA1</a>)</h6>
-                                            <p>[ Helen: Can we add a blurb here about what this is? ] The Bio-Formats API documentation is also available to read on the web <a href="#">here</a>.</p>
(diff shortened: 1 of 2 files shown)


hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/products/bio-formats/downloads/ built on staging

products/bio-formats/downloads/index.html
```diff
@@ -4,7 +4,7 @@
 ---     
         <div class="callout large primary" id="bg-image-bio-formats">
             <div class="row column text-center">
-                <h1>Bio-Formats [5.x.x] Downloads</h1>
+                <h1>Bio-Formats 5.5.0 Downloads</h1>
                 <p>[ a blurb about the Downloads ]</p>
```
hflynn: We'll need to remove this or add some text here too "Download to latest version here"?

sbesson: Yes I wasn't sure about that to put there. There is no similar template in OMERO.

lunson: @hflynn @sbesson We can always delete it unless we can put some useful note there.

sbesson: `Download the latest version` does not bring much information over the header I would say. I propose to delete for now.

hflynn: We can just delete it if you don't think having extra text there is informative/needed

hflynn:
What's the plan for previous versions? Add a 'Previous' tab and then link back to http://downloads.openmicroscopy.org/bio-formats/ ?

lunson:
@hflynn That works for me.

hflynn:
This looks fine on staging, do we want to merge it for now or were you going to add the previous tab yourself @sbesson ?

sbesson:
Assuming we are fine with this as an intermediate step, I'd capture the previous tab as a follow-up on the reference card (alongside OMERO and the other exposed software)

hflynn:
👍

PR #17, opened 2017-05-11
Enable Travis

sbesson:
This PR enables the Travis build of this repository, including a top-level `Gemfile` with the dependencies and the execution of `jekyll build`.

Note the `htmlproofer` step is currently excluded as the current branch reports 1200 broken links.


hflynn:
Merging for now since it's working, we can revisit if it raises issues later if/when we change the deployment setup.

PR #24, opened 2017-05-16
Fix favicon

sbesson:
Briefly discussed with @lunson, this PR:

- removes the duplicate favicon PNG under `img/logos`
- fix the relative link to `favicon.png` to account for `site.baseurl` and work from any location under the website.

This should also address broken links flagged in #21 

To test this PR, build the staging site, then compare

- the landing page http://snoopycrimecop.github.io/www.openmicroscopy.org/ vs https://openmicroscopy.github.io/www.openmicroscopy.org/ - should be identical
- any subpage e.g.
https://snoopycrimecop.github.io/www.openmicroscopy.org/support/ vs https://openmicroscopy.github.io/www.openmicroscopy.org/support/ - the former should now have a favicon in the tab while the latter should miss it (without this PR included)


hflynn:
👍 tested in both firefox and chrome (Safari doesn't show favicons) and it works as expected, also building locally I no longer get a favicon not found warning on the subpages whereas I still do on my branches.

PR #26, opened 2017-05-16
Add additional OMERO feature icons

lunson:
Includes icons for OME Smuggler and OME-WND-CHARM


hflynn:
Thanks Liza, those look good!

PR #27, opened 2017-05-16
Style background img text

lunson:
Unifies look and feel for all hero/callout page titles.


hflynn:
The header image on https://snoopycrimecop.github.io/www.openmicroscopy.org/products/ome-files/ still needs similar treatment too but those changes look good thanks! Are you planning to change the other header fonts too? Shall I leave this PR open for you for now so you can push further changes to this branch?

lunson:
@hflynn I pushed a fix to that page image. If all looks good on staging, this should be good to merge. Thanks!

hflynn:
👍

PR #29, opened 2017-05-18
Style Explore page contents

lunson:
Cleans up the layout for the current content


hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/explore/ is looking better thanks!

PR #33, opened 2017-05-23
Fix broken internal links

sbesson:
To review this PR, review the output of the `htmlproofer` in the Travis log, check the number of failures decreased and make sure there are no more errors of type

```
internally linking to xxx, which does not exist
```

### DIFF
--- a/about/index.html
+++ b/about/index.html
@@ -16,7 +16,7 @@ title: About Us
         <div class="row column text-center">
             <h2>The Open Microscopy Environment</h2>        
         <hr>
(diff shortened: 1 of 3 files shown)


hflynn:
So definitely improving matters 👍

hflynn:
I'll see if I can find external links for the grants and remove those links if not and I think we can probably comment out the imaging domains link (if not that entire section) as we won't have time to get more content before POTN. I'm wondering about trying a content pop-up like the team bios for the full list item.

hflynn:
I'll merge and continue on a fresh PR for the other items

PR #34, opened 2017-05-23
Fix more html proofing issues & add grant links

hflynn:
Fixing missing alt image tags and missing links

### DIFF
--- a/about/commercial-partners/index.html
+++ b/about/commercial-partners/index.html
@@ -45,7 +45,7 @@ title: Commercial Partners
         <hr> 
         <div class="row">
             <div class="medium-5 columns medium-push-6">
-                <img class="thumbnail" src="{{ site.baseurl }}/img/logos/glencoe-software.svg">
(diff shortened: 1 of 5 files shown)


hflynn:
Link fixes to follow

hflynn:
Down to only 15 failures now and @lunson is still working on the OMERO product section images so I'll leave those missing alt tags for now and fix any left over once she's done.

Main things remaining now are the full list of "Examples of OME’s resources are used in the following" for which I'm wondering if we could do a pop-up like the team bios (what do you think @lunson?) and the imaging domain link which I'll likely comment out for now as we don't have the content done.

hflynn:
Chatted to Liza briefly, I'll just add any further links needed to the existing list and delete the full list link. I just need to check where the list came from tomorrow & how many more companies there are to list.

sbesson:
Changes deployed to https://snoopycrimecop.github.io/www.openmicroscopy.org and make sense. Checked funding links manually as the `htmlproofer` is only running internally in this phase. Ready to merge.

PR #36, opened 2017-05-24
Comment out imaging domains & remove unneeded link

hflynn:
More tidying ready for the POTN preview - commented out the imaging domains section on the home page (not a priority right now to generate further content for this) and removed the unneeded link from the examples of OME resource use in external projects - when I checked back this list, we already included all the items on the current live website page so no extension was needed.

PR #39, opened 2017-05-25
Revert allowing html proofer to fail

hflynn:
As suggested on #21 this makes the html proofer failing fail Travis too now all the errors are fixed.


hflynn:
Ah, I need to merge #36 before this will pass now, closing temporarily.

PR #40, opened 2017-05-25
Update contributors list

hflynn:
A few folks have been added to https://www.openmicroscopy.org/site/about/ome-contributors since this content was copied over originally.

Staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/about/contributors/

### DIFF
--- a/about/contributors/index.html
+++ b/about/contributors/index.html
@@ -105,7 +105,7 @@ title: Contributors
             </div>
             
             <div class="large-expand columns">
-            Mark Bruce <a target="_blank" href="https://www.microvolution.com">Microvolution LLC</a>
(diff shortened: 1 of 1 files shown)


about/contributors/index.html
```diff
@@ -101,6 +105,10 @@ <h5 class="subheader">The Bio-Formats project has been fortunate to receive assi
             </div>
             
             <div class="large-expand columns">
+            Mark Bruce <a target="_blank" href="https://www.microvolution.com">Microvolution LLC</a>
```
mtbc: Should this be "Marc"? https://www.microvolution.com/company

hflynn: Oh it should indeed, it's even right on the current page!

PR #44, opened 2017-06-05
Add link to 3D viewer

jburel:
Add link to 3D viewer


mtbc:
Is this staged anywhere? I wondered if it would be at https://snoopycrimecop.github.io/www.openmicroscopy.org/explore/ and I can't see anything on http://www.openmicroscopy.org/site/support/contributing/ci-docs.html that looks like it would make it appear.

sbesson:
Should be deployed via https://ci.openmicroscopy.org/view/Mgmt/job/WWW-merge/39 (which has now be configured to be triggered daily)

PR #48, opened 2017-06-22
Community index page

hflynn:
Staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/community/

Only a few bits added here but I've merged other blurb etc so Liza could work on the styling so probably would be good for someone else to look at this at this stage and see how they think it's looking.


hflynn:
Agreed this strategy today so merging

PR #49, opened 2017-06-22
Events: create template

lunson:
Set up Events to use HTML files, not Markdown files.


hflynn:
Merging to add further content

PR #52, opened 2017-06-23
Create ome files downloads

lunson:
Creates OME Files download page, enables access to artifacts on all downloads, unify styling


products/ome-files/downloads/index.html
```diff
@@ -10,6 +10,182 @@
...
+        
+        {% include back-to-top.html %}
+        
+        <!-- begin CTA -->
+        <hr class="whitespace">
+        <header class="marketing-hero" style="background-image: url('{{ site.baseurl }}/img/bg/ome-files.jpg');">
+            <div class="row">
+                <div class="small-12 columns bg-image-text-container-cta text-center">
+                    <p class="hero-subheader">All components available for this version of OME Files are available here:</p>
+                    <a href="http://downloads.openmicroscopy.org/ome-files-cpp/" class="large btn-blue button sites-button hide-for-small-only" style="text-shadow: none;">Download Components</a>
```
sbesson: Should be https://downloads.openmicroscopy.org/ome-files-cpp/0.3.2/22/xxx where xxx is `sources` and/or `tools`. In the case of multiple subdirectories, does it make sense to have a few of these buttons?

products/bio-formats/downloads/index.html
```diff
@@ -99,67 +100,19 @@ <h3 class="text-center">Tools</h3>
...
+        
+        {% include back-to-top.html %}
+        
+        <!-- begin CTA -->
+        <hr class="whitespace">
+        <header class="marketing-hero" style="background-image: url('{{ site.baseurl }}/img/bg/bio-formats.jpg');">
+            <div class="row">
+                <div class="small-12 columns bg-image-text-container-cta text-center">
+                    <p class="hero-subheader">All components available for this version of Bio-Formats are available here:</p>
+                    <a href="http://downloads.openmicroscopy.org/bio-formats/?C=M;O=D" class="large btn-red button sites-button hide-for-small-only" style="text-shadow: none;">Download Components</a>
```
lunson: Pushed in latest commits.

lunson: I'm fine with deleting the artifact name for the OMERO ones; just say the official word :)

products/ome-files/downloads/index.html
lunson: Makes sense to have both, so I've added in latest push.

sbesson:
Thanks. Looks good to me.

PR #54, opened 2017-06-27
OMERO Downloads: add text and links

lunson:
Adds text and links to OMERO Clients section of Downloads

### DIFF
--- a/products/omero/downloads/index.html
+++ b/products/omero/downloads/index.html
@@ -13,6 +13,8 @@ title: OMERO Downloads
         <div class="row">
             <h3 class="text-center">OMERO clients</h3>
             <hr>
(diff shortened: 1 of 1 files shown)


hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/products/omero/downloads/ updated

products/omero/downloads/index.html
```diff
@@ -5,6 +5,7 @@
         <div class="callout large primary" id="bg-image-omero">
             <div class="row column text-center">
                 <h1>OMERO 5.3.2 Downloads</h1>
+                <a href="{{ site.baseurl }}/2017/05/16/omero-5-3-2.html" class="large btn-red button sites-button hide-for-small-only" style="text-shadow: none;">Read the Release Announcement</a>
```
sbesson: I like the idea of linking internally. In terms of workflow, this means that the PR bumping the version of a product will need to contain both the new release announcement as a post and an update of the product page to link to this new page.

jburel: Since we are touching this section we may want to bump the version
5.3.3 release yesterday
When it becomes the main website, we will need to adjust the card on template board (note for myself!)

PR #56, opened 2017-06-28
Move product versions and build numbers to _config.yml

sbesson:
This PR proposes to migrate the management of the product versions to the top-level `_config.yml` to ease the maintenance, updates and prevent out-of-sync issues.

The first commit leaves the artifact file sizes hardcoded in the product downloads pages for now. Happy to hear thoughts on how to deal with those. They could also be migrated to the top-level configuration file as well if we want to manage everything  that way.

### DIFF
--- a/products/ome-files/downloads/index.html
+++ b/products/ome-files/downloads/index.html
@@ -18,8 +18,8 @@ title: OME Files Downloads
                     <div class="card-section">
                         <img alt="CentOS logo" src="{{ site.baseurl }}/img/logos/3rd-party_centos.png">
                         <h5>CentOS 7.3</h5>
-                        <a href="http://downloads.openmicroscopy.org/ome-files-cpp/0.3.2/22/binaries/ome-files-standalone-bundle-0.3.2-Release-CentOS7.3-x86_64-b22.tar.xz" class="button hollow tiny"><i class="fa fa-download"></i> Release (28.28 MB)</a>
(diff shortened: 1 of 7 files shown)


products/ome-files/downloads/index.html
```diff
@@ -100,8 +100,8 @@ <h3 class="text-center">Self-contained "super" builds</h3>
                     <div class="card-section">
                         <img alt="Windows logo" src="{{ site.baseurl }}/img/logos/3rd-party_windows.png">
                         <h5>VC14 32‑bit (VS2015)</h5>
```
hflynn: While you're editing this anyway, 'Windows' is missing from the beginning of this heading

hflynn:
I'm not sure re: download sizes - that's quite a lot of extra parameters to add to the config file but on the other hand, they are easier to find there than scrolling through the html on the download page. So if we are keeping them, it might be worth migrating them.

If we move to manually updating the link to the release announcement being the only thing we need to update on the download page, that's at least near the top and easy to find.

sbesson:
Pushed the missing Windows change.

Re  artifacts sizes, I propose to postpone the migration to later. Part of the question is related to the maintenance of these values. Ideally it should be possible to either reuse/extend [ome-release](https://github.com/openmicroscopy/ome-release) and/or write a local Python script generating a list of human-readable sizes using e.g. `urllib2` as in:

```
>>> r=urllib2.urlopen("http://downloads.openmicroscopy.org/bio-formats/5.2.4/artifacts/bioformats_package.jar")
>>> float(r.headers['content-length'])/1000/1000
30.163378
>>> float(r.headers['content-length'])/1024/1024
28.76603889465332
```

sbesson:
As this has been approved and is conflicting with #69, merging so that we can move forward. As mentioned above the maintenance of download sizes will need to happen separately.

PR #59, opened 2017-06-28
news items

will-moore:
See https://trello.com/c/aTdBNblP/18-news-migration


hflynn:
I've removed all the 'Dear All' and 'Regards' bits so they don't read like emails any more and the old community page links can be updated as internal page links to ({{ site.baseurl }}/support/)

hflynn:
But I'm going to have to fix a load of old /site/ links anyway so if you just remove the 'Dear All' and 'Regards' bits, that'll save me having to fix up stuff I'm not grepping for too - thanks.

PR #69, opened 2017-06-29
optimize layout for mobile, unhide CTAs

lunson:
This optimizes the layout for mobile, and un-hides the call-to-action buttons to the Download pages (per Jason's request) for the mobile experience.

PR #70, opened 2017-06-30
Optimize layout for mobile, unhide CTAs

sbesson:
Replaces #69 with a conflict resolution

This optimizes the layout for mobile, and un-hides the call-to-action buttons to the Download pages (per Jason's request) for the mobile experience.

### DIFF
--- a/_includes/cta-omero.html
+++ b/_includes/cta-omero.html
@@ -4,7 +4,7 @@
             <div class="row">
                 <div class="small-12 columns bg-image-text-container-cta text-center">
                     <p class="hero-subheader">Get started working with your images now</p>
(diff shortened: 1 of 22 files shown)


sbesson:
Without this PR after reducing the browser size, I get 

screen shot 2017-06-30 at 16 33 12

With this PR included

screen shot 2017-06-30 at 16 33 09

Looks good to me unless there is any specific testing to be done in addition @lunson

lunson:
@seb No specific testing needed. All the Download buttons are now showing on the Product pages. I just need to un-hide the buttons from the actual Download pages, but will do so in a separate PR after this merges.

PR #72, opened 2017-06-30
OMERO Downloads: unhides CTA for mobile layout

lunson:
Small fix for mobile that un-hides a button on the OMERO Downloads page.


sbesson:
Without this PR

screen shot 2017-07-03 at 09 22 25

with this PR

screen shot 2017-07-03 at 09 22 26

:+1:

PR #75, opened 2017-07-03
Add most recent posts

jburel:
Migrate recent release posts not listed under https://trello.com/c/aTdBNblP/18-news-migration

cc @sbesson


jburel:
I have also created a new entry for figure 3.0.0 on openmicroscopy.org (so we are in synch)
This was missing.

PR #77, opened 2017-07-03
Use new permanent doc URLs

sbesson:
In preparation of tomorrow's discussion, this PR updates all the recent announcements to:

- use the new canonical documentation URLs for OMERO and Bio-Formats instead of the old `site/support` URLs
- review all the old documentation URLs to use `site/support/{omero,bio-formats}x.y` when appropriate
- drop a few `-staging` for URLs

NB: this domain is still internal while this PR is opened.


sbesson:
@jburel conflict fixed and pushed to fix all remaining OMERO/BF URLs

jburel:
Merging so we can move forward with the rest of the cleaning tasks

PR #82, opened 2017-07-09
Unify community and about menus

sbesson:
See https://trello.com/c/6l2MKwtY/28-community-about

Opening for discussion and review tomorrow


sbesson:
Discussed with @lunson @joshmoore and @pwalczysko. Closing this and we will try to use News as the top-level menu rather than Community and have the News subdirectory become Announcements instead.

PR #83, opened 2017-07-10
Top level news

lunson:
Changes Community > News, and News > Announcements as discussed with @sbesson, @joshmoore and @pwalczysko.


sbesson:
Deployed at https://snoopycrimecop.github.io/www.openmicroscopy.org/ 

See screen shot 2017-07-10 at 16 33 41

As discussed earlier the new organization works for me, allows a separation of the 2 menus and the subcategories (announcements, events,...) being now top-level works well in terms of concise URLS (www.openmicroscopy.org/announcements). 

The latter proposal raises the question on whether we should apply the same strategy to the rest of the subdirectories and flatten the structure with `products/{omero,bio-formats,ome-files}` -> `{omero,bio-formats,ome-files}` and `about/{teams, contributors,commercial-partners,licensing}` -> `{teams, contributors,commercial-partners,licensing}`. /cc @joshmoore

sbesson:
Barring vetoes, this is in agreement with what we discussed and I will follow-up with a proposal PR to flatten the hierarchy of the remaining menus (`about` and `products`)

PR #87, opened 2017-07-11
Bump Bio-Formats and OME Files

dgault:
Updates the Bio-Formats version to 5.5.3
Updates the OME Files version to 0.40
Adds missing Bio-Formats contributors from recent releases


sbesson:
Thanks, now redeployed on https://snoopycrimecop.github.io/www.openmicroscopy.org/index.html

rleigh-codelibre:
The other component version still require bumping?

sbesson:
@rleigh-codelibre: at the moment, we preserve the links to the component tags in the downloads page - mimicking the section from http://downloads.openmicroscopy.org/ome-files-cpp/0.4.0/24/#code.

If we wanted to further reduce this maintenance, we could consider reducing the versions maintained under the website to the `ome-files` and `ome-model` only and hide the detail of the individual component versions which can be inspected from the `sources` subfolder under the downloads folder. Thoughts?

sbesson:
Merging for content freeze anyways.

PR #88, opened 2017-07-11
Update more doc URLs

sbesson:
Following the migration of the last sets of Sphinx documentation under docs.openmicroscopy.org thanks to @simleo, this PR updates a set of remaining `site/support` URLs to use the versioned docs.

To test this PR check the links are correct and point to the right page.

All remaining `site/support` URLs should be pointed at more ancient documentations sets and will be updated in a follow-up PR to use www-legacy.openmicroscopy.org instead of www.openmicroscopy.org.


sbesson:
Closing in favor of #89 which also modified  `support/index.html`

PR #89, opened 2017-07-11
Add docs landing page

lunson:
Moves Docs from under Support to its own landing page. Please note that links may need to be updated.


sbesson:
Overall, I like the concept. Previously, the documentation links at the bottom of the support page was easy to miss but also felt a bit out of place given the dedicated top-level menu. My impression is that this PR provides a nice landing page for the set of technical documentations we can refer to:

screen shot 2017-07-11 at 20 31 55

Once agreed and merged, next step will be to resurrect #88 and update some of the outdated URLs in favor of the new generic ones.

sbesson:
No veto. Getting this in.

PR #92, opened 2017-07-12
Add prod configuration

sbesson:
Still under investigation, the goal of this configuration file is to override the `baseurl` used in `_config.yml` and by GitHub pages for the production deployment of this website. /cc @kennethgillen


kennethgillen:
👍 , but I don't currently have a way to test it without merging and re-running the `www-dev` playbook somewhere.

sbesson:
Tested during today's testing. All worked as expected.

PR #93, opened 2017-07-13
Doc: use new URLs consistently

sbesson:
The old legacy documentation URLs (http://www-legacy.openmicroscopy.org/site/support) will slowly be redirected towards the new docs.openmicroscopy.org URL.

In the meantime, this PR reviews the remaining legacy documentation URLs in the new website and uses the new canonical documentation URLs instead. There should be no change in content.

With this change, the only remaining occurences of hardcoded `www.openmicroscopy.org` href in the website should be 1- the forum URL in the footer, 2- the old URLs in the posts. The latter should be changed to point at `www-legacy` instead of `www` in a follow-up PR.

To test this PR, review the corresponding changes under http://snoopycrimecop.github.io/www.openmicroscopy.org/.


_posts/2017-02-21-ome-files-cpp-0-3-1.md
```diff
@@ -76,7 +76,7 @@ Note that this is a work in progress and subject to change.
 
 **Errata**
 
-See [Errata](https://www.openmicroscopy.org/site/support/ome-files-cpp/ome-cmake-superbuild/manual/html/errata.html) for this release.
+See [Errata](https://docs.openmicroscopy.org/ome-cmake-superbuild/0.3.1/errata.htm) for this release.
```
rleigh-codelibre: Also wondering, is there no automated link checking to pick this type of error up?

sbesson: :+1: will push the fix

Good point about linkchecking. We are currently disabling external linkcheck in Travis and limiting it to internal checks. I had a quick investigation at enabling it back but as we see in our daily job builds, it comes with its lot of false positives and we don't want Travis to be systematically red.

The plan for 1.0.0 was minimally for someone to run it manually, curate the output and fix the failing URLs. See https://trello.com/c/6mhivdqz/7-fix-broken-links for more information

What probably needs to happen if fine the logic in `htmlproofer` to exclude `href` or restrict domains (`*.openmicroscopy.org` would be a good start). Suggestions/help welcome.

sbesson:
Pushed the fix, I propose to handle the wrapping separately as we'll have more PRs coming against this repo to fix the content next week.

PR #94, opened 2017-07-17
Content fixes

rleigh-codelibre:
https://trello.com/c/m6JLh8pq/16-incorrect-content

### DIFF
--- a/_unused/lab-swedlow.html
+++ b/_unused/lab-swedlow.html
@@ -5,7 +5,7 @@ title: Teams
         <div class="callout large primary">
             <div class="row column text-center">
                 <h1>Swedlow Lab</h1>
-                <p>Prof Jason Swedlow, Wellcome Trust Centre for Gene Regulation &amp; Expression, University of Dundee, UK</p>
+                <p>Prof Jason Swedlow, Discovery Centre for Translational &amp; Interdisciplinary Research, University of Dundee, UK</p>
             </div>
         </div>
 
--- a/_unused/lab-swedlow.html
+++ b/_unused/lab-swedlow.html
@@ -5,7 +5,7 @@ title: Teams
         <div class="callout large primary">
             <div class="row column text-center">
                 <h1>Swedlow Lab</h1>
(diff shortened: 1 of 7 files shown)


_unused/lab-swedlow.html
```diff
@@ -5,7 +5,7 @@
         <div class="callout large primary">
             <div class="row column text-center">
                 <h1>Swedlow Lab</h1>
-                <p>Prof Jason Swedlow, Wellcome Trust Centre for Gene Regulation &amp; Expression, University of Dundee, UK</p>
+                <p>Prof Jason Swedlow, Discovery Centre for Translational &amp; Interdisciplinary Research, University of Dundee, UK</p>
```
sbesson: Although the Discover Centre is the correct name for our building, we should rather use 

```
Division of Computational Biology, School of Life Sciences, University of Dundee
```

for the affiliation across the board.

sbesson:
Changes are sensible and look good to merge. Only unknown is whether there is a proper class for highlighting the syntax under `explore/index.html`. @lunson: do you have more hindsights?

rleigh-codelibre:
@lunson Wrap or indent?  It's verbatim preformatted text, with explicit indentation which is why I also added `<pre>` tags around the `<code>`.

Could the highlighting be due to missing CSS?

lunson:
@sbesson The code class should be "html" not "sh".

rleigh-codelibre:
@lunson Why is the code class `html` when the content here is not HTML?  It's shell script; is there a class for this we could use instead?  Is there a list of the possible classes for the `<code>` element?

lunson:
@rleigh-codelibre It was just a style name that stuck. You can call the class whatever you want, but be sure to change the CSS selector in openmicroscopy.css (line 57) to match. For example, if you change it to `<code class="shell-script">`, the selector in the CSS file should be changed from `code.html` to `code.shell-script`.

rleigh-codelibre:
@lunson Ah, OK.  I was assuming this hooked into the jekyll Rouge or Markdown code formatting/highlighting stuff.

rleigh-codelibre:
Should all be fixed up now.

lunson:
Thanks @rleigh-codelibre!

PR #98, opened 2017-07-19
Convert launch of new website as announcement

sbesson:
This should turn the announcement sent to the mailing list as an announcement on the website itself.


kennethgillen:
👍

PR #103, opened 2017-08-16
Sec vul template

lunson:
Please don't merge until migration of all content is ready.

This PR includes:

- addition of Security to navigation menu
- Security landing page (for now it's just "Security Vulnerabilities" but I can expand this to include more types as we move more content)
- Sample template using the most recent copy from https://www-legacy.openmicroscopy.org/site/products/omero/secvuln/2017-SV1-filename

as discussed in https://trello.com/c/turO3BZF/64-security-vulnerabilities


security/index.html
```diff
@@ -0,0 +1,43 @@
...
+        <div id="community-posts" class="row">
+
+            <div class="small-12 medium-8 medium-offset-2 columns">
+                <h6><i class="fa fa-code-fork"></i> OMERO versions 5.2.7 and earlier</h6>
+                <h2><a href="{{ site.baseurl }}/security/2017-sv1-filename-mutability.html">2017-SV1 Filename Mutability</a></h2>
+            </div>
+            <hr class="medium-8">
+            <div class="small-12 medium-8 medium-offset-2 columns">
+                <h6><i class="fa fa-code-fork"></i> OMERO versions 5.2.7 and earlier</h6>
+                <h2><a href="#">2017-SV2 Edit in RW Group</a></h2>
```
hflynn: This filler # is breaking Travis:
- ./_site/www.openmicroscopy.org/security/index.html
  *  linking to internal hash # that does not exist (line 91)
     <a href="#">2017-SV2 Edit in RW Group</a>
htmlproofer 3.7.2 | Error:  HTML-Proofer found 1 failure!

hflynn:
Staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/security/

hflynn:
Looking nice!

sbesson:
Temporarily excluding as @hflynn is working on migrating the rest of the security content to use the new format and to allow progress on #107

security/2014-sv2-empty-passwords.html
```diff
@@ -0,0 +1,99 @@
+---
+layout: news-security
+filename: security
+title: 2014-SV1 Unicode Passwords
```
hflynn: Ah, forgot to update this when pasting the template

security/2016-sv1-cleanse.html
```diff
@@ -0,0 +1,68 @@
+---
+layout: news-security
+filename: security
+title: 2016-SV2 Share
```
hflynn: Another cut & paste I missed

sbesson:
Will perform a deeper content review this tomorrow. Glancing over the [top-level page](https://snoopycrimecop.github.io/www.openmicroscopy.org/security/), am I correct that this page is meant to list all existing vulnerabilities but also provide an answer to the question: "does my OMERO version have a known security issue?". For the latter, the headers use different conventions on versions ranges or uses of `earlier` which makes it hard to read.

lunson:
@sbesson Yes, and yes just based on existing page, but I'm open to changing it. Do you think it would be a better experience to have the user look for version first? Can we remove `and earlier` and or `later` from the wording (copied over from original content)? I can use the versions as a heading and place the secvul pages under an accordion-style list so that user is not overwhelmed with all the content at once. Plus we could avoid having to add pagination since the list won't get as long.

hflynn:
@sbesson Yeah sorry, I copied across the versions from the current page and of course the ranges chosen reflect the release version at time of writing and were never updated later, so for example the Java issue now I look at the blog post is actually 5.0.6-5.1.2 inclusive, or possibly all versions prior to 5.1.2...
We probably could change them all to 'version xx and earlier' in the headings without losing accuracy.

sbesson:
No worries, we definitely need to clarify the ranges of versions once we have the correct representation. Immediate thought while reviewing:

- `No known vulnerabilities` in 4.4.12 is probably not worth mentioning. We have dropped support for this line so we are not watching for vulnerabilities but more recent secvuln probably apply to it.
- the  Java LTS line duplication makes it hard to read as the header of the affected versions is smaller
- we probably need to homogeneize the URLs e.g. `/2017-SV1-filename` (`info`and `site/products/omero/secvuln`) vs `/2017-sv1-filename-mutability.html`
- if we wanted to have slightly shorter URLs without the trailing `.html`, we could consider using subfolder for the individual pages under a subfolder e.g. `2017-SV1-filename/index.html`. Is there a case where multiple pages for a given security would be desirable in general?
- probably outside the scope of this PR: we ship the patch for the Java LTS vulnerabilities issue since OMERO 5.1.3. However the link to the blog post seems to suggest you still need to apply it as the blog post precedes the releases. We might need to created a dedicated security page (without a SV identifier) and refer to the blog post in the background section.

On the layout question, an alternative to the accordion-style link would be something along the lines of https://www.postgresql.org/support/security/ i.e. a table listing 1- the security issues, 2- the range of known versions affected, 3- the specific version where the secvuln was fixed.

hflynn:
re: URLs - yes to making them match the old ones, sorry I missed that the first example didn't, I certainly intended for them to match. Personally, I can't see a need for more than one page so I'm not sure subfolders is worth it just to ditch .html

The Java issue probably won't need repeating if we change the version headings anyway.

Looking at that Postgresql table, it occurs to me that we could switch the version headings to the OMERO version the issue is fixed in.

I defer to @lunson on accordion list vs table!

lunson:
@hflynn Are you okay to take care of the content edits for this one?

@sbesson I'm usually opposed to tables as they're annoying to make fully responsive, but since there will only be 3 columns, let me see what I can do.

hflynn:
@lunson yes, I can do an editing sweep once we are happy with the styling

sbesson:
My personal preference goes towards https://snoopycrimecop.github.io/www.openmicroscopy.org/security2/. Assuming I have a 5.2.0 server and I want to know which vulnerabilities could affect me, the problem of the accordion list case is that it requires 3 clicks to expand the 3 lists and have the full set of vulnerabilities.

sbesson:
Staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/security/. As discussed today, this the page which reached the larger consensus in the first round of discussion. Including so that we can discuss with the rest of the team tomorrow and separately look into unifying the Affected versions column semantics.

PR #106, opened 2017-08-21
capitalize the "I" in Iviewer

lunson:
Fixes the issue in https://trello.com/c/LF6cK3DW/101-omeroiviewer-instead-of-omeroiviewer

### DIFF
--- a/omero/analyze/index.html
+++ b/omero/analyze/index.html
@@ -32,7 +32,7 @@ usergroup: omero
                 <p>Draw ROIs across multiple dimensions and measure pixel
                     intensity. </p>
                 <a class="button hollow tiny" href="http://help.openmicroscopy.org/measurement-tool.html" target="_blank">View Insight user guide</a>
-                <a class="button hollow tiny" href="https://www.youtube.com/watch?v=FQD540TTwFg" target="_blank">Watch OMERO.iviewer movie</a>
+                <a class="button hollow tiny" href="https://www.youtube.com/watch?v=FQD540TTwFg" target="_blank">Watch OMERO.<span style="text-transform: uppercase;">i</span>viewer movie</a>
             </div>
             <div class="omero-feature-container tall-content column text-center">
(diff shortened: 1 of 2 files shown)


hflynn:
it's the other way around @lunson - it's capitalized now by the styling and should be lowercase

lunson:
@hflynn Okay, I just realized that on Firefox it's lowercase and on Chrome it's not, so I thought it was the other way around. Will fix.

hflynn:
Ah, i hadn't realised that the browser made a difference, I'll be sure to check not just Safari in future!

jburel:
I was using Chrome

hflynn:
Looks fine on staging in Chrome and Firefox now. In Safari I can't actually see the difference between 'I' and 'i' on my screen which would explain how i missed this in the first place!

omero/analyze/index.html
```diff
@@ -32,7 +32,7 @@ <h2 class="header-blue">Analyze your data in OMERO</h2>
                 <p>Draw ROIs across multiple dimensions and measure pixel
                     intensity. </p>
                 <a class="button hollow tiny" href="http://help.openmicroscopy.org/measurement-tool.html" target="_blank">View Insight user guide</a>
-                <a class="button hollow tiny" href="https://www.youtube.com/watch?v=FQD540TTwFg" target="_blank">Watch OMERO.iviewer movie</a>
+                <a class="button hollow tiny" href="https://www.youtube.com/watch?v=FQD540TTwFg" target="_blank">Watch OMERO.<span style="text-transform: lowercase;">i</span>viewer movie</a>
```
joshmoore: That's one verbose "lowercase"! Would a CSS class `#productname` (or similar) be possible for usability?

lunson: It will still be "verbose" though because this is a one-off that overrides a style applied to all the buttons.

hflynn:
@jburel @joshmoore can this be merged now?

jburel:
Merging so we can move forward with the other changes

PR #110, opened 2017-08-23
Artifacts sizes generation

sbesson:
With the migration of the landing downloads page to www.openmicroscopy.org and the usage of `_config.yml` to maintain the versions, the only manual step leftover while updating downloads pages following a release remains the update of the artifact sizes.

129df8f adds a Python script which goes through all the artifacts of a released product, reads the `ContentLength` from the header opened with `urllib2` and formats it as a human-readable size using the same logic as in https://github.com/openmicroscopy/ome-release/blob/develop/doc_generator.py.

4d74113 is a manual update of the Bio-Formats downloads page using the return from `python gen_sizes.py bio-formats 5.6.0`

Open questions:
- the current request uses the default `urllib2` header e.g. `Python-urllib/2.7`. Should we update this to filter these downloads requests or is it handled via the IP filtering?
- thoughts about trying to maintain more of these parameters in `_config.yml` and expanding `gen_sizes.py` generate a YML block e.g.

```
bf:
 version: 5.6.0
 bioformats_package: 31.71 MB
 bftools: 30.03 MB
```

### DIFF
--- a/bio-formats/downloads/index.html
+++ b/bio-formats/downloads/index.html
@@ -21,7 +21,7 @@ title: Bio-Formats Downloads
                         <h5>Bio-Formats Package</h5>
                         <h6>bioformats_package.jar</h6>
                         <p class="card-caption">The complete bundle for end users containing everything needed to read images into ImageJ. Instructions for installing Bio-Formats in ImageJ are available in the <a href="https://docs.openmicroscopy.org/bio-formats/{{ site.bf.version }}/users/imagej/installing.html" taget="_blank">user guide</a>.</p>
-                        <a href="https://downloads.openmicroscopy.org/bio-formats/{{ site.bf.version }}/artifacts/bioformats_package.jar" class="button hollow small"><i class="fa fa-download"></i> Download (31.64 MB)</a>
+                        <a href="https://downloads.openmicroscopy.org/bio-formats/{{ site.bf.version }}/artifacts/bioformats_package.jar" class="button hollow small"><i class="fa fa-download"></i> Download (31.71 MB)</a>
                     </div>
                 </div>
             </div>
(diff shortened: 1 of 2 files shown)


rleigh-codelibre:
The script and changes look sensible.

For the registry/feedback DB we filter all dundee/internal IP addresses; I think we use the same rule for downloads.

Putting the information directly in YAML would be a nice improvement.  Would the hashes go there as well?

Does this need to be using urllib?  Could it run on the filesystem directly and avoid any requests?

sbesson:
If we start encoding information in the top-level yml, adding hashes as well would not be a hug deal although then we might need an extra level e.g.

```
bf:
  version: 5.6.0
  bioformats_package:
    sha1: ce99c0b06171ca1f3305f9c2277fa242ae07da6d
    size: 31.71 MB
  bftools:
    sha1: 27d8de17ab11b1fe21837aa5516a25f2724d38f5
    size: 30.03 MB
```

On using `urllib2` vs the filesystem, given the landing page is now completely decoupled from the downloads artifacts, I was assuming have access to the filesystem would not be a requirement i.e. someone outside firewall should be able to open a PR bumping the version numbers and filesizes.

An alternative here would be to repurpose the downloads jobs e.g. BIOFORMATS-DEV-release-downloads to:
1- verify the checksum on downloads.openmicroscopy.org (as it does currently)
2- in addition, generate and archive an yml block containing the information above. This block could be directly copied and pasted for opening the PR bumping the release against this repository /cc @jburel

sbesson:
Reviewing this PR, this will require some decisions and a bit of work which should really be driven by a team. Captured my representation of the status and the next potential steps on https://trello.com/c/ur2WQk6z/66-downloads-pages-checksums-and-file-sizes and closing this as I do not have the capacity to complete it.

PR #112, opened 2017-08-29
Security: versions

sbesson:
Follow up of #103, this PR reviews the affected versions column of the security table.

Amongst the various possibilities, the initial proposal is to use the same convention as https://www.postgresql.org/support/security/ i.e. describe affected versions via their `major.minor` version number rather than using ranges or `or earlier`.

Additionally this reorganizes the table to list the vulnerabilities fixed simultaneously in descending order.


mtbc:
I'd be fine with omitting the affected versions and listing only the fixed versions. Not sure that the effort of figuring when the secvuln was introduced is worth the effort and clutter.

sbesson:
On https://github.com/openmicroscopy/www.openmicroscopy.org/pull/112#issuecomment-325586451, if this makes the content of this page clearer and easier to maintain, no particular objection to dropping the `Affected versions` column in the security landing page. Happy to push an extra commit here if this helps reviewing the differences.

hflynn:
Unless we are going to ditch affected versions from the individual pages for the secvulns, it's no extra effort to have it on the index page. I like the compromise of just listing the minor versions affected here.

https://snoopycrimecop.github.io/www.openmicroscopy.org/security/index.html looks good to me

security/index.html
```diff
@@ -17,59 +17,59 @@
...
-                        <td>4.4.12</td>
+                        <td>4.4, 5.0</td>
+                        <td><a href="{{ site.baseurl }}/security/2014-sv2-empty-passwords.html">2014-SV2 Empty Passwords</a></td>
+                        <td>4.4.12, 5.0.5</td>
                     </tr>
                     <tr>
-                        <td>5.0.0 - 5.0.4, 4.x - 4.4.11</td>
-                        <td><a href="{{ site.baseurl }}/security/2014-sv2-empty-passwords.html">2014-SV2 Empty Passwords</a></td>
-                        <td>4.4.12</td>
+                        <td>4.4,5.0</td>
```
hflynn: No space between versions here

sbesson:
@hflynn leaving you to sign off on this PR. Coming back to https://github.com/openmicroscopy/www.openmicroscopy.org/pull/103#issuecomment-324607086, while starting to look at the redirects from www-legacy and the info URLs, I realized there is an obvious benefit in using the same naming scheme for these URLS with a different namespace (i.e. ``site/products/omero/secvuln` -> `/security`), I'll open the proposed renaming as a follow-up PR

hflynn:
All looks good on staging

PR #113, opened 2017-08-29
Fix secvuln heading

hflynn:
Spotted when reviewing https://snoopycrimecop.github.io/www.openmicroscopy.org/security/index.html#secvuln-reporting


sbesson:
Thanks. And just realized that maybe we should also fix the typo in the page name (missing `t` in `authenication`)...

hflynn:
👍 but i realise you'll need to do that on your PR because we'll conflict otherwise

hflynn:
Closing Pr and letting @sbesson cherrypick my commits to #112 instead

PR #115, opened 2017-08-30
Security URLs

sbesson:
Follow up of https://github.com/openmicroscopy/www.openmicroscopy.org/pull/112, this PR:

- migrates the individual security pages under subfolders. With this change, the URL of individual security pages should be e.g. https://www.openmicroscopy.org/security/2012-SV1-ldap-authentication/ and match the current naming of the `info/vulnerabilities` and `/site/products/omero/secvuln` URLs and simplify the redirect work
- prefix all versions with `OMERO` since the security pages are no longer under the `omero` namespace


sbesson:
Staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/security/

hflynn:
Filenames now match the old plone URLs 👍 

I think having 'OMERO' prefixing each version number when we have two listed is overkill - "OMERO 4.4.12, 5.0.5" reads fine

hflynn:
Looks good to me 👍

joshmoore:
General :+1: for reduction in `/foo.html` in favor of `/foo/` (Likely coming to idr.o.org as well)

PR #119, opened 2017-09-04
Security advisories

sbesson:
Follow-up of #103 and #112, this PR reorganizes the security section as follows:

- `security/` is now the landing page for all security related content
- `security/advisories` is the landing page for the security advisories
- `security/advisories/YYYY-SVn-xxxx` is the landing page for a given advisory

See also:
- https://jenkins.io/security/advisories/
- https://www.mozilla.org/en-US/security/advisories/
- https://www.vmware.com/uk/security/advisories.html


sbesson:
See https://snoopycrimecop.github.io/www.openmicroscopy.org/security/

security/index.html
```diff
@@ -1,82 +1,14 @@
 ---
 layout: news
-title: Security Vulnerabilities
+title: Security
 filename: security
-main-blurb: Known vulnerabilities, workarounds, and resolutions
```
hflynn: We probably want a new blurb here
"Keeping your data secure" ?

sbesson: Probably. I can push `Keeping your data secure` and leave others to open follow-up PRs here and in the following section.

sbesson:
@hflynn: pushed another commit which renames the layout as `security-advisory` and fixes the link to link back to the advisories list and updates the main security page description.

I propose not to modify the content of the `/security/advisories/` page here as my goal is to get the final URLs in place for live deployment. We can discuss and update the content of this page as follow-up PRs. In addition to removing the `Affected versions` column, another consideration for this page might be to add another link `back to Security`.

hflynn:
Yeah, I think a 'Back to Security' link will be needed too. But happy for that change to be in a follow up.

URLs all implemented as suggested so good to merge from me

chris-allan:
My only request would be to restore the advisory dates to the table on the `/security/advisories/` page. Probably as an additional column.

sbesson:
@chris-allan: :+1: the first column of the advisories page has now been updated accordingly.

PR #125, opened 2017-09-08
Add previous to download footers &fix OMERO components link

hflynn:
See https://trello.com/c/ALSwsNNV/91-add-previous-navigation

This adds a link to the downloads index to the bottom of each downloads page below the full components button.

I also fixed the OMERO components button link which wasn't using the auto version replacement so was therefore still pointing at 5.3.2.


hflynn:
Staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/omero/downloads/ etc

PR #134, opened 2017-09-25
indicate current page of pagination

lunson:
Adds clearer visual indicator of current page in pagination of Announcements; cleans up margins


hflynn:
Tested on Safari, Firefox and Chrome - all showing current page with white text in a blue box so nice and clear 👍

hflynn:
Got a thumbs up from Jason on the Trello card so merging

PR #138, opened 2017-09-28
Add sitemap

lunson:
Adds sitemap of current content to footer


hflynn:
Staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/site-map/

will-moore:
Could the OME Files heading to go under Bio-Formats instead of being on the same line since it is really the 3rd item in Products and it gets a bit lost on the right.
I see now that elsewhere on the page you've got siblings going across the page, so I guess it's not "wrong" but just appeared different from the other Projects because OMERO is on it's own row.

The Explore heading looks like somethings missing since it's not obvious that the headings are also links so it looks like there's actually no pages in the Explore section of the site. It needs a link, e.g. "Explore" in the same way as the "Overview" link under Projects.

The [site map] link on page footers looks a bit isolated, but don't know if there's a better place for it? so I guess it's OK.

lunson:
@will-moore All set. 

I think the [ Site Map ] looks isolated too but not sure where to put it; I typically see it with a "Terms of Service" but for now, it's going to fly solo.

Thanks for feedback!

hflynn:
Did you want to add the new 'On the Web' page to this too @lunson ?

hflynn:
Looks good to me, merging for deployment with the 5.4.0 release

PR #139, opened 2017-09-29
Remove Simone from the Dundee team page

hflynn:
Since Simone has left the Dundee team and gone back to CRS4

### DIFF
--- a/teams/index.html
+++ b/teams/index.html
@@ -169,23 +169,6 @@ title: Teams
                     </div>
                 </div>
                 <div class="column">
-                    <a class="team-member" data-open="modal9">
-                        <img class="thumbnail" alt="photo of Simone Leo" src="{{ site.baseurl }}/img/team/simone_leo.jpg">
-                        <h5>Simone Leo</h5>
-                        <p class="job-title">Software Developer</p>
-                        <span class="bio-link tiny button hollow">Read Bio</span>
-                    </a>                                                            
-                    <div class="reveal" id="modal9" data-reveal>
-                      <h4 class="text-center">Simone Leo</h4>
-                      <p class="job-title">Software Developer</p>
-                      <hr>
-                      <p>Simone Leo joined the Dundee team in 2015 after being a satellite developer for OME based at CRS4 in Italy. He received an M.Sc. degree in Electronic Engineering from the University of Cagliari, Italy, in 2002 and a PhD in Biomedical Engineering from the same university in 2015. Before joining CRS4 in 2006, he worked in the private sector as a system administrator. His research interests are mainly focused on distributed computing models and applications, with an emphasis on their applicability to computational biology.</p>
-                      <button class="close-button" data-close aria-label="Close reveal" type="button">
-                        <span aria-hidden="true">&times;</span>
-                      </button>
-                    </div>
-                </div>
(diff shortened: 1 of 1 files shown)


hflynn:
Staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/teams/

sbesson:
Makes sense although to me this raises the question on whether we should think of a location for listing previous team members. It feels a bit unfortunate that we maintain a list of external contributors to the project but people who used to be in the org and contributed a lot fully disappear when they move forward.

hflynn:
That is a good point. Maybe we could add a former team members section to the contributors page?

hflynn:
Started adding everyone to the contributors page then realized that adding non-devs here was a bit tricky (the page specifically says 'direct contributions to the software') but it seemed really unfair to list everyone except Kelli (and Joyce) so I've moved the list to the bottom of the team page instead so we can acknowledge everyone.

hflynn:
Staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/teams/

hflynn:
Content from https://www-legacy.openmicroscopy.org/site/about/development-teams/jason

teams/index.html
```diff
@@ -169,23 +169,6 @@ <h4 class="text-center">Roger Leigh</h4>
...
-                      <p class="job-title">Software Developer</p>
-                      <hr>
-                      <p>Simone Leo joined the Dundee team in 2015 after being a satellite developer for OME based at CRS4 in Italy. He received an M.Sc. degree in Electronic Engineering from the University of Cagliari, Italy, in 2002 and a PhD in Biomedical Engineering from the same university in 2015. Before joining CRS4 in 2006, he worked in the private sector as a system administrator. His research interests are mainly focused on distributed computing models and applications, with an emphasis on their applicability to computational biology.</p>
-                      <button class="close-button" data-close aria-label="Close reveal" type="button">
-                        <span aria-hidden="true">&times;</span>
-                      </button>
-                    </div>
-                </div>
-                <div class="column">
                     <a class="team-member" data-open="modal10">
```
mtbc: It's okay that these "modal" `data-open` attribute values are no longer contiguously numbered?

hflynn: AFAIK it's fine as long as the ID matches between the team member section and the reveal section. Is that correct @lunson ?

mtbc: That makes sense, I can test on that basis.

teams/index.html
```diff
@@ -403,4 +386,84 @@ <h4 class="text-center">Wilma Woudenberg</h4>
         <hr class="whitespace">
         <!-- end  Development teams -->
         
+         <!-- begin Former team members -->
+            <hr class="whitespace">
+            <div class="row text-center">
+                <div class="small-12 columns">
+                    <h3>Former members of the OME team in Dundee</h3>
```
mtbc: "in Dundee": so might Harald someday move down here?

hflynn: Well, strictly it's paid by Uni of Dundee since Seb isn't based here physically either, but I'm not sure we need to be that nitpicky for this context!

mtbc: Okay. :smiley:

PR #140, opened 2017-09-29
Links to docs

lunson:
Adds links to Documentation from both the Product pages and their Downloads pages

### DIFF
--- a/ome-files/downloads/index.html
+++ b/ome-files/downloads/index.html
@@ -5,6 +5,7 @@ title: OME Files Downloads
         <div class="callout large primary" id="bg-image-ome-files">
             <div class="row column text-center">
                 <h1>OME Files Downloads</h1>
+                <a href="https://docs.openmicroscopy.org/latest/ome-files/" target="_blank" class="hero-link">Read the Docs</a>
             </div>
         </div>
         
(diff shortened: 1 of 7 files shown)


ome-files/downloads/index.html
```diff
@@ -5,6 +5,7 @@
         <div class="callout large primary" id="bg-image-ome-files">
             <div class="row column text-center">
                 <h1>OME Files Downloads</h1>
+                <a href="https://docs.openmicroscopy.org/latest/ome-files/" target="_blank" class="hero-link">Read the Docs</a>
```
hflynn: This link doesn't actually work cc @sbesson

hflynn:
@lunson this is going to conflict again as soon as I open the PR for the 5.4.0 release so you may as well hold off fixing it until next week.

When you do so, can you also change the broken ome-files doc link to use the current version replacement for now since the latest redirect isn't working at the moment. Thanks!

lunson:
@hflynn Okay, will put this on hold.

hflynn:
@lunson you'll be able to merge in master and fix this now, thanks

ome-files/downloads/index.html
sbesson: We definitely need to sort out the documentation/API URLs for the upcoming OME Files 0.5.0 release and unify with the rest of the project. I will start a document listing the various elements so that we can discuss on a concrete proposal. For now, a workaround would be to add the documentation bundle to the downloads page which contains a hierarchical version of all the documentation.

hflynn:
I'll take this over so we can at least get a temporary fix for the ome files docs URL which is broken on the live site.
Thanks @lunson

lunson:
@hflynn Thanks, Helen. Sorry I didn't get to it sooner.

hflynn:
No problem! Seb and I were chatting about the best temp fix for OME Files so it seemed easier to just sort it rather than relaying the decision back to you, especially when I realized that my suggested fix in earlier comments wouldn't work either!

PR #145, opened 2017-10-09
Use qa link for demo account on omero downloads

hflynn:
See https://trello.com/c/9i6FPk0G/20-downloads-information

cc @jburel


hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/omero/downloads/ updated

jburel:
Thanks 
works as expected.
I think we should include it for the release 
cc @sbesson

PR #148, opened 2017-10-11
Docs links

hflynn:
Replaces #140 

Adds more docs links for all products and uses a temporary fix the OME Files docs - the current OME Files docs at docs.oo.org are ONLY the C++ docs, the full bundle is only hosted on www-legacy so far or bundled in the download. A proper fix needs putting in place for 0.5.0, which @sbesson is on the case of but for now better to point at a legacy URL that has all the content than a 404 or only part of the docs.


hflynn:
No more docs or downloads links broken anyway and aside from a couple of links that are genuinely timing out when I tested them, all the other failures are just flakiness. Just investigating whether we can set the checker to just check docs.oo.org and downloads.oo.org before I give up and disable the external checker again completely.

hflynn:
Travis doesn't like the option from the htmlproofer docs for some reason so I've just disabled the external checker again so I can get this built on snoopy for review.

hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/ome-files/ etc updated for review

sbesson:
:+1: LGTM and deployed as this is an obvious 404 in our website.

PR #150, opened 2017-10-16
Use https for twitter links

hflynn:
See https://trello.com/c/Vl41pXd4/120-on-the-web-page-should-use-https-outbound-links

### DIFF
--- a/on-the-web/index.html
+++ b/on-the-web/index.html
@@ -14,30 +14,30 @@ title: On the Web
                 <h3 class="social-media-feature-section-headline"><i class="fa fa-twitter" aria-hidden="true"></i> Follow Us on Twitter</h3>                
                 <div class="social-media-feature-section-inner">                
                     <div class="social-media-feature-section-feature top-left">
-                        <a class="account-profile" href="http://www.twitter.com/openmicroscopy" target="_blank"><img class="account-profile-pic" src="https://pbs.twimg.com/profile_images/430266361785229313/MU4Sx0sn_400x400.png" alt="@openmicroscopy twitter profile pic"></a>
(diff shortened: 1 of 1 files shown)


hflynn:
Staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/on-the-web/

on-the-web/index.html
```diff
@@ -14,30 +14,30 @@
...
                         <div>
-                            <p class="account-handle"><a href="http://www.twitter.com/bioformats" target="_blank">@bioformats</a></p>
+                            <p class="account-handle"><a href="https://www.twitter.com/bioformats" target="_blank">@bioformats</a></p>
                             <p class="account-desc">Development status feed of coming features and release announcements.</p>
                         </div>
                     </div>                        
                     <div class="social-media-feature-section-feature bottom-left">
-                        <a href="http://www.twitter.com/idrnews" target="_blank"><img class="account-profile-pic" src="https://pbs.twimg.com/profile_images/847712210067398657/IZywWoxH_400x400.jpg" alt="@idrnews twitter profile pic"></a>
+                        <a href="https://www.twitter.com/idrnews" target="_blank"><img class="account-profile-pic" src="https://pbs.twimg.com/profile_images/847712210067398657/IZywWoxH_400x400.jpg" alt="@idrnews twitter profile pic"></a>
```
mtbc: Interesting that the IDR anchors don't get the `class="account-profile"` like the other two.

hflynn: Huh, i hadn't even noticed that! It doesn't seem to make any difference to the rendered page but I could add them in for completeness. Is there a reason behind this @lunson ?

lunson: @hflynn The right margin was off for those two so I left the class out, but if you're not seeing any difference, then yes, feel free to add the class.

hflynn: Sorry, I meant the fact that they are missing makes no difference to how they look compared with the others. If it made the margin off, we can certainly leave it as is.

PR #151, opened 2017-10-17
host social media images

lunson:
See https://trello.com/c/zz6XsOqQ/119-twitter-profile-images-arent-shown-in-firefoxs-private-window


hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/on-the-web/ updated

hflynn:
Tested in a private Firefox window and can confirm, without this PR the live site doesn't show the icons (as per Simon's screenshot) but with this PR the snoopy site now displays the icons as expected. Merging, thanks.

PR #153, opened 2017-11-10
Use docker-compose text on Explore page

hflynn:
See https://trello.com/c/xea4BDu4/104-publicize-docker-compose

Staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/explore/

### DIFF
--- a/explore/index.html
+++ b/explore/index.html
@@ -61,12 +61,24 @@ EOF</code></pre>
             </div>
             <div class="medium-4 column end">
                 <h5 class="text-center">Start a Server in Docker</h5>
-                <pre><code class="html"># Start a server in docker
-docker run -d --name postgres -e POSTGRES_PASSWORD=postgres postgres
-docker run -d --name omero-master --link postgres:db \
-    -e DBUSER=postgres -e DBPASS=postgres -e DBNAME=postgres \
-    -p 4063:4063 -p 4064:4064 \
-    openmicroscopy/omero-grid master
+                <pre><code class="html">version: '3'
+
+services:
+  db:
+    image: postgres
+    environment:
+      - POSTGRES_PASSWORD=postgres
+  omero:
+    image: openmicroscopy/omero-server:5.4
+    environment:
+      - CONFIG_omero_db_user=postgres
+      - CONFIG_omero_db_name=postgres
+      - CONFIG_omero_db_pass=postgres
+      - ROOTPASS=omero
+  web:
+    image: openmicroscopy/omero-web-standalone:master
+    ports:
+      - 4080:4080
 </code></pre>
             </div>
         </div>


joshmoore:
Thanks, @hflynn. Would be interesting to hear if @manics and @kennethgillen saw anything similar at devops days. I'm still looking for the right place to store the compose.yml. A gist would be easy enough but it seems too unofficial.

cc: @sbesson @jburel

kennethgillen:
Not I, sorry. @manics?

joshmoore:
@hflynn / @sbesson : would an include from somewhere be possible? Or would it have to be via code-generation then? Or use this repo's travis?

manics:
Isn't a link sufficient?

joshmoore:
Not really. The goal is to have something that can be seen on this page that a user could reproduce, ideally via cut-n-paste. We could drop `compose` in favor if the 3 `docker run` commands (would still need testing though)

manics:
I don't see the point of a cut-and-paste `docker-compose.yml` file, since you need to create the file to use docker-compose.

joshmoore:
Clicks count in terms of getting people to understand what we're selling them, but agreed, needing to create the file is suboptimal.

hflynn:
The Jekyll docs suggest to me that we can only include content that is within this repo. We can put the content in a file of its own if that helps with using Travis to test it though.

joshmoore:
Works for me. Any objections?

sbesson:
So the proposal would be to use this repo as the current placeholder for a reference `docker-compose.yml` file and have it responsible for the systematic testing? This would work for me as I do not think we have a better place at the moment - see https://github.com/openmicroscopy/omero-web-docker/issues/10#issuecomment-341145003.

A few open questions:

- do we want to amend Travis to test the `docker-compose.yml` workflow in addition to the Jekyll build or in a different Matrix build? if so, should we also switch to cron jobs to report issues early?
- in terms of the release workflow, while OMERO patch releases should not need updates, there might be implications for minor/major releases of OMERO since the announcement/downloads page might have to go public before we update the omero-{server,web} Docker images which would lead in a two-step update of this repository

joshmoore:
* responsible for testing: yes
 * matrix or not: don't really mind either way
 * cron: :+1:
 * major releases: I'd assume we'd manage the version in the compose file itself (rather than using 'latest'), no?

manics:
Does that mean all production docker-compose files will go in this repo?

joshmoore:
> Does that mean all production docker-compose files will go in this repo?

What other production docker-compose files do you foresee?

manics:
Variants of recommended production configs e.g. public user, high-performance configs with redis and multiple webs/dbs, microservices, web-apps, the new queuing system.

hflynn:
If we are planning a whole raft of these files, a separate repo may be the best way to go after all, they are not going to be very discoverable in the /_includes/ directory here.

joshmoore:
NB: https://github.com/docker/compose/issues/1818

manics:
If you want something prepackaged might as well use kubernetes with helm then

joshmoore:
I think that's getting ahead of ourselves. I haven't even _seen_ a helm chart for OMERO yet. (Once we have a chart and especially once kubernetes is integrated into docker, then this certainly sounds like a nice to have.)

@hflynn: if we can have a file in a separate repository still be shown under "explore", then I say let's give that a try. If we _can't_, then let's extract the file to an include so that we can test it (in another PR), and still get this up on the website. At the moment, we suggest people *not* use what's currently on the live website, so while we're figuring out the whole package, let's at least get that replaced with something that works today.

manics:
Is this a bad time to point out the `# Start a server with ansible` is broken in two ways?
- the playbook uses out of date omero-server and postgres roles
- the `cat playbook.yml <<EOF` uses incorrect syntax

manics:
(see https://github.com/ome/ansible-examples-omero for tested versions)

hflynn:
@manics if you can give me the text for the ansible section, I can fix that too - either that or could you open a PR please?

manics:
If it's just a link to that repo I can write some text. If you want an embedded script similar to what's already there that's more work.

joshmoore:
:+1: This works for me as an immediate fix for the deprecated command. I can review the testing based on the discussion in gh-155. (Perhaps eventually we'll need/want a carousel with these)

cc: @lunson

lunson:
@hflynn I'm having trouble getting what's staged so I'll style this after merge.

hflynn:
Shouldn't be any styling needed for this at the moment @lunson, it's the 'Start a Server in Docker' section on https://snoopycrimecop.github.io/www.openmicroscopy.org/explore/ - the content is rendered the same as the live page.

lunson:
@hflynn I see, but the 'OME Developers' section as a whole just needs tidying up after whatever PR is making that big gap appear after the Ansible part :)

hflynn:
The Ansible section isn't sorted yet, Josh wants to use the same approach as here for #155 too

joshmoore:
i.e. styling after this and 155?

lunson:
@joshmoore I'll hold off on styling until after they're both merged.

joshmoore:
Cheers. So translating myself from https://github.com/openmicroscopy/www.openmicroscopy.org/pull/153#issuecomment-345264092, merging.

PR #155, opened 2017-11-16
Replace ansible example with link to repo

manics:
The ansible example has been broken since it was first added to the explore page. This replaces it with a link to a working examples repo. In future we can reconsider whether to embed the example.

See https://github.com/openmicroscopy/www.openmicroscopy.org/pull/153#issuecomment-344257148

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ install: bundle install
 script:
     - bundle exec jekyll build  -d _site/www.openmicroscopy.org
     - bundle exec htmlproofer ./_site --disable-external
-    - cd _includes; docker build -t omero-c7 .
-    - docker run --privileged -v /sys/fs/cgroup:/sys/fs/cgroup:ro -d --name omero-c7 omero-c7
-    - docker exec omero-c7 bash /tmp/omero-c7.sh
-    - echo TODO - add docker-compose test after merge
+    #- cd _includes; docker build -t omero-c7 .
+    #- docker run --privileged -v /sys/fs/cgroup:/sys/fs/cgroup:ro -d --name omero-c7 omero-c7
+    #- docker exec omero-c7 bash /tmp/omero-c7.sh
+    #- echo TODO - add docker-compose test after merge
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,17 @@
 language: ruby
-sudo: false
+sudo: required
 cache: bundler
 rvm: 2.3.3
 env:
   global:
   - NOKOGIRI_USE_SYSTEM_LIBRARIES=true # speeds up installation of html-proofer
+services:
+  - docker
 install: bundle install
 script:
     - bundle exec jekyll build  -d _site/www.openmicroscopy.org
     - bundle exec htmlproofer ./_site --disable-external
+    - cd _includes; docker build -t omero-c7 .
+    - docker run --privileged -v /sys/fs/cgroup:/sys/fs/cgroup:ro -d --name omero-c7 omero-c7
+    - docker exec omero-c7 bash /tmp/omero-c7.sh
+    - echo TODO - add docker-compose test after merge
--- a/_includes/Dockerfile
+++ b/_includes/Dockerfile
@@ -1,0 +1,3 @@
+FROM centos/systemd
+COPY omero-c7.sh /tmp/
+CMD ["/usr/sbin/init"]
--- a/_includes/omero-c7.sh
+++ b/_includes/omero-c7.sh
@@ -1,0 +1,23 @@
+#!/bin/sh
+yum install -y -q epel-release
+yum install -y -q ansible sudo
+ansible-galaxy install \
+    openmicroscopy.postgresql \
+    openmicroscopy.omero-server \
+    openmicroscopy.omero-web
+cat << EOF > ~/omero-playbook.yml
+- hosts: localhost
+  roles:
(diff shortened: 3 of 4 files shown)


joshmoore:
I'd vote for embedding the working example here & now. I'll offer to get travis testing both the ansible and the docker-compose files.

joshmoore:
@manics : can you pick a [MWE](https://en.wikipedia.org/wiki/Minimal_Working_Example) and paste it here?

lunson:
This incorporates changes from https://github.com/openmicroscopy/www.openmicroscopy.org/pull/153#issuecomment-344257148 so I'll get merge conflicts if I open a PR against Simon's branch, but here's the styling: https://github.com/lunson/www.openmicroscopy.org/tree/style-explore-ansible

hflynn:
@manics Josh is still waiting for action from you on this.

manics:
`curl https://gist.githubusercontent.com/manics/b534bfeade3075d571fd5e5ec7b1a5b7/raw/f99c35d6de4f0f1ee576bbd5ef8c7d3e59a0ff33/omero-c7.sh | sh`

joshmoore:
For purposes of this discussion, here's the old text:

```
# Start a server with ansible
ansible-galaxy install openmicroscopy.omero-server
cat playbook.yml <<EOF
- hosts: localhost
  roles:
  - role: openmicroscopy.omero-server
    postgresql_users_databases:
    - user: omero
      password: omero
      databases: [omero]
EOF
```

@manics text:

```
#!/bin/sh
yum install -y -q epel-release
yum install -y -q ansible sudo
ansible-galaxy install openmicroscopy.postgresql openmicroscopy.omero-server openmicroscopy.omero-web
cat << EOF > ~/omero-playbook.yml
- hosts: localhost
  roles:
    - role: openmicroscopy.postgresql
      postgresql_databases:
      - name: omero
      postgresql_users:
      - user: omero
        password: omero
        databases: [omero]
      postgresql_version: "9.6"
    - role: openmicroscopy.omero-server
      omero_server_rootpassword: ChangeMe
    - role: openmicroscopy.omero-web
EOF
ansible-playbook ~/omero-playbook.yml

```

which is executable (:+1:) but assumes CentOS (:-1:). An alternative would be:

```
# playbook.yml - roles can be installed from https://galaxy.ansible.com/openmicroscopy
- hosts: localhost
  roles:
    - role: openmicroscopy.postgresql
      postgresql_databases:
      - name: omero
      postgresql_users:
      - user: omero
        password: omero
        databases: [omero]
      postgresql_version: "9.6"
    - role: openmicroscopy.omero-server
      omero_server_rootpassword: ChangeMe
    - role: openmicroscopy.omero-web
```
though the link to galaxy could be outside the `<code/>` block.

manics:
The roles are only tested on CentOS 7

joshmoore:
True. Which means we should add that to the header "Deploy an OMERO server on CentOS 7 with Ansible"

manics:
What's the final decision on this?

joshmoore:
Travis is testing the `sh` based version. If we decide that's the route we want to go, then perhaps the same should be done on the compose side:

```
#!/bin/sh
pip install docker-compose
cat << EOF > ~/docker-compose.yml
...
EOF
docker-compose up
```

joshmoore:
Doh. Helped travis along with systemd, but now failing on:

```
RUNNING HANDLER [openmicroscopy.omero-server : omero-server restart omero-server] ***

fatal: [localhost]: FAILED! => {"changed": false, "msg": "Unable to start service omero-server: Failed to start omero-server.service: Unit not found.\n"}

	to retry, use: --limit @/root/omero-playbook.retry
```

I'll dig more tomorrow.

joshmoore:
I've failed to get this passing in travis. I'm going to rollback to a non-tested state, update for the PG 3.0 roll, and see if we can get it into the 5.4.2 release.

kennethgillen:
@joshmoore @manics: What testing do you require before inclusion on the site? 

Manual run of omero-c7.sh in docker?

joshmoore:
Yes.

joshmoore:
Testing in an `centos/7` vagrant:

```
mkdir /tmp/c7
cd /tmp/c7
vagrant init centos/7
vagrant up
vagrant ssh
curl ... > script.sh
sudo bash script.sh
```
succeeds for me.

Having tested the script locally, a few items perhaps for (later?) discussion since they are all at the cost of complicating the nice and tidy script:

 * Add `set -e` since if the user runs this as `bash /tmp/script.sh` rather than `sudo bash /tmp/script.sh` it runs several commands rather than failing immediately.
 * Put the playbook in a less destructive location. (Conceivably this could overwrite a user's file.)

hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/explore/ updated 

This could do with some styling changes - you have subheading size font rather than the smaller text used in the following sections.

joshmoore:
Can we perhaps handle that as well as a block of text for the docker column to the right in a follow-on since we're somewhat abusing @manics' PR here?

hflynn:
That works for me

hflynn:
Are we merging this for release as an improvement over the current content and tidying later? Happy to do so, the styling nitpicks are minor.

joshmoore:
I'd say let's get this in and the minor fix on top of it as well. I noted it on release cards farther to the right so I wouldn't forget, but happy for you to run with it.

PR #163, opened 2017-12-04
 Update ome-files docs links to new URLs

hflynn:
With the 0.5.0 release, the docs.oo.org URL now serves the index page allowing users to navigate to all the docs sets from the single URL so we don't need the old /site/support/ URLs any more.


hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/ome-files/ etc updated

PR #170, opened 2017-12-13
fix link under Attaching Analytical Results

lunson:
Minor fix from https://trello.com/c/izVlz266/131-fix-link-for-attaching-analytical-results

### DIFF
--- a/omero/analyze/index.html
+++ b/omero/analyze/index.html
@@ -24,7 +24,7 @@ usergroup: omero
                 <h4>Attaching analytical results</h4>
                 <p>Support for storing analytical results attached to
                     images.</p>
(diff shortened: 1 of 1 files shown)


hflynn:
Actually it was intended to point at http://help.openmicroscopy.org/managing-data.html#attach but this new link is good for HCS data. Could we have both - the managing data link as "view attaching files guide" and the new one as "view guide for HCS data" ?

hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/omero/analyze/ looks good to me, happy @will-moore ?

will-moore:
Looks good, thanks

PR #171, opened 2017-12-13
replace old JCB logo with new one

lunson:
Minor fix: replacing old JCB logo https://trello.com/c/tyiJ1iR3/132-add-new-branding-for-jcb


hflynn:
👍 looks good thanks

PR #172, opened 2017-12-13
Fix docs links

lunson:
See https://trello.com/c/hkvSttvR/127-docs-page-has-unnecessary-read-on-the-web-links

### DIFF
--- a/css/openmicroscopy.css
+++ b/css/openmicroscopy.css
@@ -445,6 +445,9 @@ div.medium-3 > .card-download { min-height: 491px; }
 a.card-as-link > .card-support:hover {
     border: 1px solid #03a9f4;
 }
+    a.card-as-link > div > div > p.card-caption {
+        height: 148px;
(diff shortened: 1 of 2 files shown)


lunson:
Closing PR. I realized I targeted the wrong tag.

lunson:
Fixed HTML structure.

hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/docs/ updated - the boxes aren't a consistent size which looks a bit odd - especially for OME Files where the box ends up being the biggest despite it having the smallest amount of text

hflynn:
Of course now I realise the boxes are the same on the live page, I just never noticed before!

hflynn:
Looks a lot better now thanks - they still aren't exactly the same but it's a lot less glaring now.

Happy with this @manics ?

manics:
That's a lot better 👍

PR #179, opened 2018-01-12
Update gdoc meeting notes link for 2018

hflynn:
The old link points at the 2017 folder so this updates to point at the new 2018 one.

I'd prefer to point at the top folder so people can access 2017 and 2018 notes, but I can't work out how to make the gdoc permissions work for that


hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/on-the-web/ updated

hflynn:
Discussed at web meeting today - permissions are a nightmare so let's stick with pointing at 2018 folder. We can maybe archive the 2017 notes under downloads as an additional future solution for anyone really interested - just stick a zip of everything.

PR #183, opened 2018-01-25
add meta tags for Twitter

lunson:
Allows logo preview when ome.org is linked in a tweet

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -10,6 +10,10 @@
         <meta property="og:url" content="{{ site.baseurl }}" />
         <meta property="og:title" content="The Open Microscopy Environment" />
         <meta property="og:description" content="A consortium of universities, research labs, industry and developers producing open-source software and format standards for microscopy data." />
+        <meta name="twitter:title" content="The Open Microscopy Environment">
+        <meta name="twitter:description" content="A consortium of universities, research labs, industry and developers producing open-source software and format standards for microscopy data.">
+        <meta name="twitter:image" content="{{ site.baseurl }}">
+        <meta name="twitter:card" content="{{ site.baseurl }}/img/logos/ome-logo_fb.png">
         <title>{{ page.title }} | Open Microscopy Environment (OME)</title>
         <link rel="shortcut icon" type="image/x-icon"
               href="{{ site.baseurl }}/img/favicon.png">
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -10,10 +10,10 @@
         <meta property="og:url" content="{{ site.baseurl }}" />
         <meta property="og:title" content="The Open Microscopy Environment" />
         <meta property="og:description" content="A consortium of universities, research labs, industry and developers producing open-source software and format standards for microscopy data." />
+        <meta name="twitter:card" content="summary">
         <meta name="twitter:title" content="The Open Microscopy Environment">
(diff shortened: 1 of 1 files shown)


hflynn:
I can't seem to test this using the snoopy site so we might have to just merge it and see if it works!

hflynn:
Sadly implementing part of this on the blog doesn't seem to have worked:

blog_twitter_preview

hflynn:
Ah, I shouldn't have removed the 'title' section for the blog - https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/summary tells me it is required.

So no reason why this shouldn't work here.

hflynn:
Not having any joy getting this working on the blog this afternoon but I was maybe too impatient and didn't give the merged code time to filter through to the live site. I'll retest the blog update tomorrow and if it works then, we'll know this should work too.

_includes/head.html
```diff
@@ -10,6 +10,10 @@
         <meta property="og:url" content="{{ site.baseurl }}" />
         <meta property="og:title" content="The Open Microscopy Environment" />
         <meta property="og:description" content="A consortium of universities, research labs, industry and developers producing open-source software and format standards for microscopy data." />
+        <meta name="twitter:title" content="The Open Microscopy Environment">
+        <meta name="twitter:description" content="A consortium of universities, research labs, industry and developers producing open-source software and format standards for microscopy data.">
+        <meta name="twitter:image" content="{{ site.baseurl }}">
+        <meta name="twitter:card" content="{{ site.baseurl }}/img/logos/ome-logo_fb.png">
```
hflynn: This image location needs to be under "image" and the content of the "card" has to be "summary"

Not sure if the order matters too, but https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/summary has the "card" item first

hflynn:
Finally got this to work for the blog on https://github.com/ome/blog/pull/55 - you need to use the full URL for the image too

_includes/head.html
hflynn: Also need to full URL not the site.baseurl code

lunson: Okay, I've pushed changes in the next commit.

hflynn:
Changes look good, this is now the same as what's working on the blog. I'm not entirely sure what this logo will look like cropped to a square but maybe we can just change it later if it's not pretty.

@joshmoore ok to merge for deployment today with the 5.4.2 retraction?

joshmoore:
WFM

hflynn:
This is what it looks like:

website_preview

Don't know if we are bothered by the cropped logo or not. I do wonder if we should remove the description so that the field is autofilled by text from the actual page though.

lunson:
I'll work on a square logo and leave the description blank. I'll open a separate PR.

hflynn:
Thanks @lunson

hflynn:
Also, a heads-up, it only works for *https* - don't know if there is anything we can do about that

lunson:
@hflynn Do you mean use https when linking to the logo for the image meta tag?

hflynn:
Sorry, I mean the image didn't work when I forgot we default to https and posted a link to http://openmicroscopy.org instead.

hflynn:
I should really remember to use https though so not a big deal!

PR #190, opened 2018-01-29
figure 3.2.0 announcement

will-moore:
Ported from https://github.com/ome/omero-figure/pull/261

### DIFF
--- a/_posts/2018-01-30-figure-3-2-0.md
+++ b/_posts/2018-01-30-figure-3-2-0.md
@@ -13,7 +13,7 @@ New features:
  - Rename of channel rendering setting "Reverse intensity" to "Inverted"
 
 These features are included in the updated <a href="http://help.openmicroscopy.org/figure.html">User Manual</a>
-and a new <a href="https://www.youtube.com/watch?v=mjlNBUoGKRc&">demo video</a>.
+and a new <a href="https://www.youtube.com/watch?v=mjlNBUoGKRc">demo video</a>.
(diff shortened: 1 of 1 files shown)


hflynn:
Rebuilding the snoopy staging site now

hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/2018/01/30/figure-3-2-0.html looks good

jburel:
Will it possible to open help and video in a new tab?

hflynn:
Ah yeah, good idea @jburel - that is the policy we use elsewhere on the site for external links.

will-moore:
None of the other announcements open another tab for links (just checked the last few).
Also I think it's generally a bad idea to open links in a new tab for a webpage (I guess it's OK for apps like gdoc or trello) but e.g. github doesn't do it.
e.g. https://www.smashingmagazine.com/2008/07/should-links-open-in-new-windows/

hflynn:
I give @jburel final say, I copied Liza's implementation elsewhere on the site but default to leaving users to open their own tab in general so obviously forgot for the announcements!

jburel:
in our we go to different websites
that's what we have done for example here
https://snoopycrimecop.github.io/www.openmicroscopy.org/omero/publish/
click on the video and it opens a new tab

will-moore:
Fixed links to open in new tab.

hflynn:
👍

PR #192, opened 2018-02-01
optimize OMERO pages navigation

lunson:
Not ready for merge. Just need to draw attention to https://trello.com/c/IRGd8elG/141-optimize-omero-pages-navigation in order to prepare for the OMERO.figure and iviewer landing pages. Thanks!

PR #193, opened 2018-02-05
Optimize OMERO navigation

hflynn:
Replaces #192 

@lunson this is working now so i'm opening this PR but feel free to fetch the branch and add to it if that was your plan. Otherwise, we could probably have this go live as is after some review, and then add the Plugin & Figure stuff in a further PR.

### DIFF
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -6,6 +6,17 @@
     <script src="{{ site.baseurl }}/js/responsive-tables.js"></script>
     <script>
         $(document).foundation();
+        try {
+            var paths = document.location.pathname.split('/').filter(
+                function(el) { return el.trim() !== ''});
+            var path = paths[paths.length-1];
+            $(".menu-subnav li a").each(function(i, el) {
+                if (el.href.indexOf(path) !== -1) {
+                    $(el).addClass('is-active');
+                    return false;
+                }
+            });
+        } catch(ignored) {}
     </script>
     <script>
       (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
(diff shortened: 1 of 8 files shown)


lunson:
I'm fine with it going live apart from the Plugins/Figure pages, but it only works at the top level and isn't working at the `layouts/omero-features.html` level yet. Thanks, @waxenegger!

_includes/scripts.html
```diff
@@ -6,6 +6,17 @@
     <script src="{{ site.baseurl }}/js/responsive-tables.js"></script>
     <script>
         $(document).foundation();
+        try {
+            var paths = document.location.pathname.split('/').filter(
+                function(el) { return el.trim() !== ''});
+            var path = paths[paths.length-1];
+            $(".menu-subnav li a").each(function(i, el) {
```
hflynn: @waxenegger can we just add "header-subnav" here somehow to make this work for the omero-features tabs menu too?

waxenegger:
@hflynn I can offer this: https://github.com/waxenegger/www.openmicroscopy.org/commit/5561580f36617de19e4b15980eb6ff273a832cab

As is, this doesn't 'scale' indefinitely, i.e. in terms of hierarchy it only includes the header-nav, so one parent. The whole crux with this active setting is that anything is a page as opposed to a single page app. That's perfectly good, in fact that's what I'd expect given the nature of the site.
It does mean that I have no state for navigation though and I have to use the url in order to know where I am. For this to work for a parent (i.e. features + import) I had to put it into the NEW directory so that the url will reflect that NEW is its parent. That way I know I have to put import and NEW active.

So that's something to keep in mind for the future. 

2 notes should you want to use the changes: 

- links to the folders I moved might need updating in other places (checked but might have missed something) 

- what's under "institutions" will need to modified work like "developer", "features" and "scientists": sub header bit in _layouts folder with same html for navigation  (ul - li -a tags) as well as class name ('header-subnav'), then it should work like the others.

hflynn:
Thanks @waxenegger 

We need to decide whether the improvement is worth changing URLs for so I'm going to open another alternative PR with your proposal in it.

hflynn:
Temporarily excluding so we can build #195

hflynn:
Closing this as the alternative has been approved now

PR #194, opened 2018-02-05
Add OMERO.figure landing page

hflynn:
See https://trello.com/c/U9PIYgHx/40-figureopenmicroscopyorg-site-migration

This page would replace the current figure.openmicroscopy.org site and can be updated to point to new release announcements and videos for each release.

Navigation is not in place yet but this will lie within a Plugins section which will be added to the new navigation in #193


hflynn:
Staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/omero/figure/

joshmoore:
What are all the planned links _to_ this page?

hflynn:
I'm thinking we put the plugin index page up as part of the new navigation menu proposed on #193 and link from there and from the /omero/publish page, which I'm currently redrafting.

omero/figure/index.html
```diff
@@ -0,0 +1,156 @@
...
+            </div>
+            <div class="figure-feature column">
+                <i class="icon-fa-blue fa fa-crop fa-2x"></i>
+                <h5>Crop to ROI</h5>
+                <p>Use Regions of Interest from the OMERO server to crop images.</p>
+            </div>
+            <div class="figure-feature column">
+                <i class="icon-fa-blue fa fa-signal fa-2x"></i>
+                <h5>Accurate scalebars</h5>
+                <p>Use the pixel size metadata from OMERO to add 'live' scalebars that update as you zoom.
```
mtbc: Should close this with `</p>`.

will-moore:
Looks great!
I think the tagline "Create fast figures from your OMERO images" doesn't quite work. It almost sounds like the figures themselves are fast, not the creation of them?
Maybe something like "Easily create figures from your OMERO images" or "Create quality figures from your OMERO images"? Probably the latter is better - emphasis quality over speed.

It would be nice to get the demo replaced by public figures on nightshade or demo server before we take down figure.openmicroscopy.org since it's been quite widely used.

Most of the non-release blog posts are mostly my personal dev notes or are duplicated elsewhere in some form. I'll probably port the ones I care about to http://will-moore.github.io/.
However, I think this one is still worth keeping available: http://figure.openmicroscopy.org/2015/04/30/quality-figures-from-omero-figure.html
If there's nowhere to keep it on www.openmicroscopy.org then maybe we could link to it once I've move it elsewhere?

hflynn:
@will-moore I'll update the tagline.

There isn't a definite timeline for turning off the old site anyway - we need to get the new OMERO navigation working and the links to the old site won't all be updated until 5.4.4 at the earliest anyway - so I think it's fine to leave it until we sort demo.

I think we could quite happily re-post that blog post on the main blog - I'll open a PR to do that and let you deal with the rest as you see fit.

hflynn:
Also note that if the alternative navigation proposal is approved, this page will need to be moved to /omero/plugins/figure/index.html

will-moore:
Looks good, thanks

hflynn:
Holding off merging while we decide on the navigation issue and get the plugins index page sorted.

hflynn:
Marking this as on hold for now - may be easiest to have one PR with this and the new plugins landing page, structured for the new OMERO navigation system once that is merged.

hflynn:
So I think we have three options here:

1. Leave this page as is in the structure, link from the publish and scientists sections and add it to th site map under OMERO at the same level as scientists etc.

2. Move this page under scientists, alongside 'Getting started' etc as a 4th tab. This would require some regigging to the content as this page already has its own header.

3. Wait until we've settled on an Add-ons or whatever section and then move this below that in the navigation.

will-moore:
I really like the current figure header, and the url  /omero/figure/ makes sense too. Trying to remember /omero/scientists/figure/ would be more painful / less tweetable etc.
So my preference is for option 1.

hflynn:
Further discussion still favouring option 1, especially as that means we keep the nice individual header.

hflynn:
Pushed an update to the figure version and added some crosslinks so we have a full implementation of option 1 to look at.

hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/omero/features/publish/ and https://snoopycrimecop.github.io/www.openmicroscopy.org/omero/scientists/ updated to include links to the new page and https://snoopycrimecop.github.io/www.openmicroscopy.org/omero/figure/ now links to the latest release announcement (3.2.1)

joshmoore:
Minor comment:

screen shot 2018-02-20 at 08 39 43

I wonder if the `Help` link could be removed when/if the Figure page did that linking itself. That would get more people more quickly to the more nicer content.

hflynn:
I can do that now, the figure page links to the help. I removed the help link in favour of the new page on the scientists section but not the feature page so actually your suggestion makes it more consistent too.

hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/omero/features/publish/ updated and link on https://snoopycrimecop.github.io/www.openmicroscopy.org/omero/figure/ tweaked to make clear you're going to the user guide not the OMERO docs.

hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/site-map/ updated to include figure page too.

This PR's current state reflects a complete implementation of option 1 ready for discussion at the next website meeting

joshmoore:
Few comments going over this with others:

 * few cases of `OMERO.Figure`
 * can the new menu bars be added to `omero/figure` even if there's no dedicated highlighted button

screen shot 2018-02-23 at 11 36 08
screen shot 2018-02-23 at 11 36 01

hflynn:
Ah i forgot about the CSS capitalising words, I'll fix those.

I _think_ I can add the OMERO menu, I'll see if I can get it working this afternoon.

hflynn:
All implemented now @joshmoore 

Interestingly I missed the OMERO.Figure bits because Safari doesn't interpret the CSS capitalization as applying to .xxx words and i hadn't looked at the staging site on Chrome until I saw your screenshots and was confused!

joshmoore:
All 3 changes look good. (Good to know that adding the menu bar is so straight-forward). :+1:

hflynn:
I did get some odd errors in my local build when clicking around but travis is passing and the staging site works fine so I'm assuming we are all good to add the menu even though the highlighting can't work because the page isn't in its hierarchy.

hflynn:
I think this is ready to go once @sbesson & @jburel give their thumbs up so removing the on-hold label

jburel:
It looks good
@joshmoore did you have a chance to look at the view count on yourtube after migrating a video?
It will be nice to have the video under the org

joshmoore:
I found fairly authoritative comments that it's _not_ possible to migrate a video. Every comment I saw was "download and re-upload".

jburel:
ok so in that case, until we do new videos we will need to keep them as they are otherwise we will use the number of views

hflynn:
Merging this at EOB unless @sbesson vetoes this afternoon

PR #195, opened 2018-02-07
Alternative proposal for OMERO navigation

hflynn:
This gets the active menu highlighting working at both levels of the new OMERO submenus but does so by moving all the features pages down a level in the hierarchy so that rather than /omero/view etc they are now /omero/features/view, and the 'NEW' page is just /omero/features/index

This will likely require redirects to be put in place AND updates to the docs and help - I'll investigate and list the changes needed for the latter.


hflynn:
Only remaining quirk is that the submenu index pages (Scientists Summary etc and Features NEW) all show as underneath the 'Overview' header menu item - I'm guessing there is nothing to be done about this, I don't think it's too confusing but others may find it so possibly.

Changes now shown on https://snoopycrimecop.github.io/www.openmicroscopy.org/omero/

waxenegger:
@hflynn I'll have a look at that.

waxenegger:
@hflynn let me know if  https://github.com/waxenegger/www.openmicroscopy.org/commit/8b8e69e1ce0215176e69066cfadf73bc0de8e87a does away with the confusion :wink:

hflynn:
Thanks @waxenegger - i couldn't cherry-pick your commit for some reason but copying what you'd done did it indeed fix the issue :)

hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/omero/ updated - all now works as intended so we just need to decide if we're prepared to break live URLs. There'd be the 9 feature section URLs to redirect.

I'll check Help and the docs for links that would also need updating if this was deployed.

hflynn:
No links to the features section from any of the Help pages but there is a link to the features section in the docs globalomerotoc which would need updating so deploying this would be dependent on a docs release (most likely the 5.4.4 release). I'll open the PR for that now so I don't forget about it.

hflynn:
Other than the linked docs PR, this would just require the 9 live URLs redirecting

hflynn:
@lunson One thing we have noticed is that these new menus on the header disappear if you make the page narrow enough

lunson:
@hflynn Here's the menu style for mobile screen size: https://github.com/lunson/www.openmicroscopy.org/tree/style-alt-nav

hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/omero/institution/ updated @joshmoore @sbesson

Gemfile
```diff
@@ -25,3 +25,5 @@ end
 gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
 
 gem "html-proofer"
+
+gem "jekyll-redirect-from"
```
sbesson: Locally, using the same workflow as the deployment playbook, this gem (and the paginate one) need to be put under the `jekyll_plugins` group to work.

hflynn:
Hopefully fixed by new commit, I can't test locally as I can't get jekyll to work without deleting my local Gemfile :(

hflynn:
Finally got my local environment working properly, thanks @sbesson - this should be okay now

joshmoore:
NB: this is looking quite nice.

hflynn:
👍

sbesson:
Deployed on live website.

PR #200, opened 2018-02-21
Rename 2018 user meeting

hflynn:
June informs me that we aren't using '13th Annual' this year and the event should be called '2018 OME  Annual Users Meeting instead.


hflynn:
Staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/2018/02/09/user-meeting-2018.html https://snoopycrimecop.github.io/www.openmicroscopy.org/events/13th-annual-users-meeting-2018.html and https://snoopycrimecop.github.io/www.openmicroscopy.org/events/

PR #202, opened 2018-02-23
Chrome bg fix

lunson:
This is a fix for the background images on the Users Meeting banner not showing up in Chrome for some users.

This converts SVG background images to PNGs so this will work cross-browser.

Downsides:
- Images won't be as crisp as SVGs are for high-res screens
- If Chrome won't render the gradient sky, then users will see a light blue sky instead so not exactly the same view


hflynn:
I'm not seeing any difference between this and the live site so looks good to me. @joshmoore may like to test too given he was having issues on Friday and it was all working fine for me by then.

joshmoore:
Refreshing several times on https://snoopycrimecop.github.io/www.openmicroscopy.org/ always shows the image.

PR #205, opened 2018-02-27
Add Artwork page to share logos

hflynn:
See https://trello.com/c/FpxHM0Aw/70-artwork

Shall I link to the svg files in this repo for people to download? Or do we want to upload the logos to downloads.openmicroscopy.org? Do we want to include another format and have a zip bundle?


hflynn:
Staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/artwork/ and https://snoopycrimecop.github.io/www.openmicroscopy.org/site-map/ - looks like I'll need to do something about the display size, although my local build looked fine, they are displaying bigger and in need of some whitespace borders on the snoopy staging site.

hflynn:
Now I realise we'll obviously need at least one zip bundle because we have one link for all the product logos.

hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/artwork/ updated

joshmoore:
Minor suggestions:

 * "Download Products Logos *& Marks*"
 *

hflynn:
the problem with adding all the product marks is that it messes up the layout of the page - I tried playing with it but couldn't get it to work nicely so this may require some further styling from @lunson

hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/artwork/ updated with a rather hacked layout - it looks okay to me but I won't be offended if you want to restyle the page @lunson ;)

joshmoore:
I certainly prefer this to the previous. I'd be happy to go with it, and we can add improvements.

lunson:
Great job, @hflynn!

sbesson:
I assume the next step following live deployment might be to add a redirect for http://www-legacy.openmicroscopy.org/site/art as well?

hflynn:
Yes, I'll open a PR for that now @sbesson

PR #210, opened 2018-03-16
Add page title for features section

hflynn:
See https://trello.com/c/cSY3Ftr4/146-odd-page-name - I'd forgotten to add page titles for this section.

At the moment this PR adds "OMERO Features" to all the subpages but in the scientists and devs sections, there are different titles for the different tabs (e.g. 'Getting Started') so the alternative is to add individual titles - "New with OMERO, "Import with OMERO" etc. Happy to change to that strategy if preferred.

### DIFF
--- a/omero/developers/docs.html
+++ b/omero/developers/docs.html
@@ -1,6 +1,6 @@
 ---
 layout: omero-usergroup-devs
-title: Documentation
+title: OMERO for Developers
 usergroup: developers
 --- 
 
--- a/omero/developers/getting-started.html
+++ b/omero/developers/getting-started.html
@@ -1,6 +1,6 @@
 ---
 layout: omero-usergroup-devs
-title: Getting Started
+title: OMERO for Developers
 usergroup: developers
 ---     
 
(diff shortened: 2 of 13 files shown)


hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/omero/features/ etc updated

joshmoore:
I can also see going the other way (option 2) so that the other areas (e.g. "Scientists") also stay more consistent:

screen shot 2018-03-19 at 08 29 49
screen shot 2018-03-19 at 08 29 42
screen shot 2018-03-19 at 08 29 35

And for completeness, option 3 would be to have the whole shebang? e.g. `$AREA | $SUBAREA | OME`

hflynn:
I'm not sure if we can have subtitles, also people likely have too many tabs open to view that long a title, I'd have thought?

I'll change the scientists and devs ones to match instead - I think that makes the most sense.

hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/omero/scientists/getting-started.html etc updated now

will-moore:
Looks fine and makes sense to me.

PR #213, opened 2018-03-19
Remove link redirected to legacy for old meetings

hflynn:
As discussed at today's web meeting, this removes the link which is being redirected to legacy for the old meeting content


hflynn:
Grepping this is the only link to the /site/ URLs that isn't a docs URL being redirected to docs.openmicroscopy.org separately.

PR #219, opened 2018-04-02
Add training section

hflynn:
WIP


hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/training/ now in place - I just haven't got the active menu option highlighting working yet

hflynn:
Lates commit gets the active menu option highlighting to work but does so by moving the submenu to the bottom of the header.

@waxenegger is it possible for your active submenu script to work for a menu that is in a bar below the header? If so, can you help us out here because I can't get it working.

If not, @lunson this is going to require a bit of CSS restyling, because the colours are too dark to show up well against the header image.

hflynn:
For reference, this is what Liza originally styled:
preferred_layout

hflynn:
https://snoopycrimecop.github.io/www.openmicroscopy.org/training/ reflects the current state of this PR

waxenegger:
@hflynn sorry, you have caught me a bit out of context. so as is it's working, right? But you'd like to change the style/html?

If so, the whole script solution is quite specific in a sense that it operates on the url/path to be able to determine what's active. Also, syntactically, the options need to have a `.header-subnav` (`.menu-subnav` for its children, if any) and the list element structure they do. Should they deviate from the structure in that the html elements should be different (say something other than lists) we'd need to identify them by classes, rather than the type of element, in which case I'd have to change the script. Bear in mind that there are limits. I have to rely on a certain pattern/common denominator for this to work. If you want me to look into that I'd like to point me to the actual html layout you are envisioning. I really need to see the code to try that out. 

And yes, if the structure and position of the menu is right but you just want different looks, then forcing different looks on it by means of css is the better option. `.header-subnav` (`.menu-subnav` for its children, if any) has to remain however, you'd have to override with a second class or a style attribute then.

hflynn:
@waxenegger if you look at the _layout/training.html changes I made in https://github.com/openmicroscopy/www.openmicroscopy.org/pull/219/commits/16c607340128432deebf67391b54d7680722038b you'll see how Liza wanted it to work and what I did.

If you could possibly fix it so we can have the submenu in a separate bar below the header, that would be really helpful - my 'fix' doesn't really scale, especially as we may want to add additional text above the submenu bar.

waxenegger:
@hflynn the problem with the script is not so much any additional text so long as it is outside the ordered list (ul element) that is expected. The older solution (commit you mentioned) doesn't have that list structure and the script's selectors at the moment expect them. I can make that a bit more flexible in terms of freedom of elements to use and location by creating separate classes for that purpose which is the cleaner solution and has the pleasant side-effect that the classes used would not have css/style associated with them, and shouldn't be assigned style. But I'd have to touch all the pages that have a menu to do so.

waxenegger:
@hflynn just a follow-up to show that you can place things outside the header tag without affecting the workings of the script. I realize it's far from ideal and requires css overrides.

image

Note that the only thing that's necessary for the script is the `<ul class="header-subnav"><li><a>.....</ul>`

Here's the entire code, please don't use it as is :wink:, I have just very quickly put it together:
```
---
layout: default
title: Training
---

<header class="header" id="bg-image-training" style="margin-bottom: 0rem;">
    <h1 class="headline">Training</h1>
    <p>These materials are available for you to use and adapt for your own training purposes.</p>
</header>

                <div style="position: relative; top:50px;">
                    <ul class="header-subnav media-options-bar row small-12 medium-6 medium-offset-3"
                        style="background: gray; max-width: 100%">
                        <li>
                            <a href="{{ site.baseurl }}/training/">
                                <button class="training-option">
                                    <i class="fa fa-file-o" aria-hidden="true"></i>
                                    <span class="media-type">Docs</span>
                                </button>
                            </a>
                        </li>
                        <li>
                            <a href="{{ site.baseurl }}/training/slides/">
                                <button class="training-option">
                                    <i class="fa fa-film" aria-hidden="true"></i>
                                    <span class="media-type">Slides</span>
                                </button></a>
                        </li>
                        <li>
                            <a href="{{ site.baseurl }}/training/videos/">
                                <button class="training-option">
                                    <i class="fa fa-video-camera" aria-hidden="true"></i>
                                    <span class="media-type">Videos</span>
                                </button>
                            </a>
                        </li>
                    </ul>
            </div>
        {{ content }}
        <hr class="whitespace">
```

waxenegger:
Here's a commit that would use custom classes for the active setting: https://github.com/waxenegger/www.openmicroscopy.org/commit/d2f81c2b6bb9832412bb240036ec56275c64432c

hflynn:
Ah! I thought it had to be instead the header tag! Thanks @waxenegger I think i should be able to get this working as Liza intended on Monday now

hflynn:
@waxenegger I can't seem to get this to work for the life of me and I've don't understand why :( I've got bullet points and no active highlighting - is it just because I'm using the CSS classes @lunson has set?? 

Sorry if I'm being really dense here - it seems like I have all the elements required for the script but maybe there's something in the CSS that's upsetting it and when you said "requires css overrides" above, you meant "won't work with the existing CSS".

waxenegger:
@hflynn yes, sorry, I should have checked that. I did change the 'mechanics' of the script based on new classes but neglected to check the `is-active` which is only ever used together with the old classes:

https://github.com/openmicroscopy/www.openmicroscopy.org/blob/master/css/openmicroscopy.css#L215
and:
https://github.com/openmicroscopy/www.openmicroscopy.org/blob/master/css/openmicroscopy.css#L236

So you are now going to have to define a css rule for `is-active` such that the training page will show a certain active style. I'd not recommend using the classes I introduced for the script purpose so that they don't get a style pinned on them.

I'm not sure how you'd want the menu to look but for the purpose of applying an 'is-active' style adding this to the css is a possible way:
```
.training-subnav-bar li a.is-active button {
    background-color: black;
}
```

I'm making use of `.training-subnav-bar` which is something that you may or may not want to do depending on whether you like to keep it (for its css) or not.

waxenegger:
Forgot to say...given the nature of the menu and the changed script a solution that does not depend on  `<ul><li>` any more the following is also possible:
```
<!-- begin Training Resources navigation -->
<div class="training-subnav-bar-container">
    <div class="custom-nav training-subnav-bar row small-12 medium-6 medium-offset-3">
            <a class="custom-nav-option" href="{{ site.baseurl }}/training/">
                <button class="training-option">
                    <i class="fa fa-file-o" aria-hidden="true"></i>
                    <span class="training-type">Docs</span>
                </button>
            </a>
            <a class="custom-nav-option" href="{{ site.baseurl }}/training/slides/">
                <button class="training-option">
                    <i class="fa fa-film" aria-hidden="true"></i>
                    <span class="training-type">Slides</span>
                </button>
            </a>
            <a class="custom-nav-option" href="{{ site.baseurl }}/training/videos/">
                <button class="training-option">
                    <i class="fa fa-video-camera" aria-hidden="true"></i>
                    <span class="training-type">Videos</span>
                </button>
            </a>
    </div>
</div>
<!-- end Training Resources navigation -->
```
together with the following css the highlighting should work: 
```
.training-subnav-bar a.is-active button {
    background-color: black;
}
```

jburel:
Now the various training repositories are taking shape. The layout envision few months back is probably not going to work.
We should discuss that on Monday.

hflynn:
Okay, I'll hold off any further work on this then @jburel 

Thanks anyway @waxenegger !

lunson:
FYI - I have to interview a candidate for a couple of hours on Monday morning so I will miss our web meeting. Please tag me on any notes for the layout of this so I can figure out what to design. Thanks!

hflynn:
Looks like we'll just want a single page giving an overview of how you can use the materials and linking out to event pages and training materials on GitHub and on our downloads site. @jburel is going to start drafting some content in a gdoc and I'll add to it.

I'm wondering if it might be easier to just start again from scratch given the extensive history of this PR! I think the only thing we'll want to keep is the header and the new page and top menu entry.

lunson:
I'm fine with starting from scratch in terms of designing the layout to accommodate the content.

jburel:
@lunson 
Hopefully we should be able to re-use some style from existing pages
and take from there. That will be a good learning exercise for us

lunson:
Sounds good, @jburel. Ping if you need me.

PR #225, opened 2018-05-07
Remove URLs to registration URL for 2018 Users meeting

sbesson:
The only leftover place is the post from February 23th which we could edit or leave in place.

### DIFF
--- a/events/13th-annual-users-meeting-2018.html
+++ b/events/13th-annual-users-meeting-2018.html
@@ -10,14 +10,7 @@ main-blurb: The 2018 OME Annual Users Meeting will be held at the University of 
         <hr class="whitespace">
         <div class="row">        
             <div class="column">
-                <h2 class="event-day">Registration is now OPEN!</h2>
-                <p class="event-day-description">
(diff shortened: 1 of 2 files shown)


kennethgillen:
Would you leave "Join us May 30th to June 1st in Dundee!" if there's no link to registration?
screen shot 2018-05-07 at 12 59 35

kennethgillen:
I'd think it make sense to keep the date clearly visible, so LGTM. Though is Internal/OME slack:  https://openmicroscopy.slack.com/archives/C0K5EED3R/p1525692075000224 destined for here?

sbesson:
Deployed live.

kennethgillen:
👍

jrswedlow:
Banner should point to the meeting page: https://www.openmicroscopy.org/events/13th-annual-users-meeting-2018.html

sbesson:
The removal of the `See the schedule` link was an oversight. It should be restored in https://github.com/openmicroscopy/www.openmicroscopy.org/pull/223 alongside the addition of the programme

PR #231, opened 2018-05-29
OME2018: add references to presentations

sbesson:
To be updated as talks and workshops get published.


sbesson:
I have linked to everything was accessible currently on downloads.openmicroscopy.org. My vote would be for merging and deploying this and adding new hyperlinks as we copy content.

Agreed on the idea of thinking about  landing pages for workshops composed on multiple components.

PR #233, opened 2018-06-15
docker: Use branch name rather than "master"

joshmoore:
The automated master branch is not regularly tested
and should be removed once this change has been propagated
to the website.

### DIFF
--- a/_includes/docker-compose.yml
+++ b/_includes/docker-compose.yml
@@ -13,6 +13,6 @@ services:
       - CONFIG_omero_db_pass=postgres
       - ROOTPASS=omero
   web:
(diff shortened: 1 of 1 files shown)


joshmoore:
Thanks. Any suggestions on a deployment timeline? Or shall I just go ahead and roll it out?

sbesson:
No current planned timeline. Happy to have you go ahead and roll it out. Otherwise, we could have this routinely done as part of the weekly SA meeting until we invest into a more automated solution.

joshmoore:
Deployed live: https://www.openmicroscopy.org/explore/

PR #240, opened 2018-07-12
Include a version/date string in footer

manics:
The [new deployment process](https://github.com/openmicroscopy/www.openmicroscopy.org/pull/239) is so quick it'd be nice to have confirmation that the site was updated. This adds the build (tag) date to the footer.
screen shot 2018-07-12 at 16 19 42

Note that since this uses the build date and not the tag it's possible these will disagree, for example if you do a manual build, or if travis is delayed to the day after the tag


sbesson:
No objections to getting this in as it is. Having some form of versioning for the build time in the footer sounds like a valuable improvement. Since you changed the title to WIP, are you expecting to push more?

manics:
It was to make it clear it wasn't to go in for the 5.4.7 release

PR #242, opened 2018-07-25
Include future-dated posts in build

manics:
Whilst testing a related build from a separate private repo we discovered `future: false` inhibits the generation of future-dated posts. Since the [new deployment workflow decouples the build from the deployment](https://github.com/openmicroscopy/www.openmicroscopy.org/pull/239) this means you can't tag the repo and run the build ahead of time, it has to be done on the same day as the deployment. This PR therefore removes this flag.

### DIFF
--- a/_prod.yml
+++ b/_prod.yml
@@ -1,4 +1,2 @@
 baseurl: ''
 ga_token: 'UA-49953125-6'
-
-future: false


jburel:
The change makes sense.
I removed the flag locally while preparing 5.4.7 
and it works as expected

sbesson:
Thanks @manics. Agreed the workflow changes is allowing build and deployment dates to be fully decoupled. Options are:

- keep the current strategy and only tag this repository i.e. build a production archive on the release date. Both GitHub Pages and local builds using `config.yml` should allow to review future posts
- remove `future: false` as proposed here. No immediate objection but we had disallowed this explicitly in https://github.com/openmicroscopy/www.openmicroscopy.org/commit/3655297185f826f0ab50ac74de98ad34aa293615#diff-644fc8d4b14c6114fe62ecfa394d0309 but I cannot remember the rationale.

sbesson:
Merging to test https://github.com/openmicroscopy/www.openmicroscopy.org/pull/246

PR #244, opened 2018-08-07
Next workshop

jburel:
Update front page to promote next workshop

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,9 +6,9 @@ title: Home
         <header class="marketing-hero" id="bg-image-users-meeting">
             <div class="row homepage text-center">
                 <div class="small-12 columns bg-image-text-container">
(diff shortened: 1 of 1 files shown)


jburel:
We can certainly combine them.
Pushing my commit to @will-moore's PR now

jburel:
Closing. Pushed commit to https://github.com/openmicroscopy/www.openmicroscopy.org/pull/243

will-moore:
Do we want to remove the Dundee image ```header id="bg-image-users-meeting"``` too and revert so microscopy image as before?
Then the change would be more noticeable too.

jburel:
I don't think it is "Dundee" specific. Plane and submarine mean "travel/on the road" so I thought it was still legitimate to keep it

PR #251, opened 2018-08-28
have security advisories mention assigned CVEs

mtbc:
Staged under https://snoopycrimecop.github.io/www.openmicroscopy.org/security/advisories/.


joshmoore:
Links all work. Other than than the 3 newer CVEs state:

```
AWAITING ANALYSIS
This vulnerability is currently awaiting analysis.
```

:+1:

PR #245, opened 2018-08-09
training pages

will-moore:
See https://trello.com/c/2VSqMQmj/18-website-training-page-update

Ready for first look at these pages.
Added link to training page from "About Us" menu and support page.

Will be staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/training/
once build is run.

cc @jburel @pwalczysko

PR #258, opened 2018-09-06
No top links

will-moore:
This is an attempt to fix the issue of the top menu items being links (as well as toggling submenus) as discussed in #245.
It means that the pages linked to by the top links can be missed. Also, they are not reachable on mobile-sized screen, since they *only* toggle the submenu when on a small screen and don't behave as links.

With this PR, the main landing pages are linked to with an "Overview" link that is the first link in each sub-menu.
It hasn't been possible to remove the ```<a>``` links since they are used by foundation.js to toggle the mobile links menu, but we can disable their default link behaviour with ```preventDefault()```.

To test, try clicking top-links and their submenu items with large screen and small screen (can just resize your browser to phone-size).

### DIFF
--- a/_includes/navbar-main.html
+++ b/_includes/navbar-main.html
@@ -12,6 +12,7 @@
                     <li class="show-for-small-only"><a href="{{ site.baseurl }}/index.html"><img src="{{ site.baseurl }}/img/logos/ome-logomark.svg" alt="OME logo" style="height:32px; width:32px;"/></a></li>
                     <li class="has-submenu"><a href="{{ site.baseurl }}/about/">About Us</a>
                         <ul class="submenu menu vertical" data-submenu>
+                            <li><a href="{{ site.baseurl }}/about/">Overview</a></li>
                             <li><a href="{{ site.baseurl }}/teams/">OME Teams</a></li>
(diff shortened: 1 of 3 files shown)


sbesson:
I can certainly confirm that on mobile the current top links are not available. In addition to the `support` page now affected by the addition of the `training` page, other pages like https://www.openmicroscopy.org/about/ suffer from the same limitation.

From my perspective, the solution chosen here i.e. creating an `Overview` top-level item for the landing page of each submenu is reasonable. It also matches what we implemented for the OMERO pages under https://www.openmicroscopy.org/omero/  where the `OVERVIEW` item  navigation bar  brings you back to the landing OMERO page.

Main concern/question is an UX question about whether we want to disable the top-level links at the same time as introducing the Overview item or whether we want to do that later on. For mobile users this is effectively a no-op so the decision only affects large screen users.

joshmoore:
> Main concern/question is an UX question about whether we want to disable the top-level links at the same time as introducing the Overview item or whether we want to do that later on. For mobile users this is effectively a no-op so the decision only affects large screen users.

As they are, the top-level changes color on mouseover making me thing they are links.

will-moore:
Top links don't change colour or cursor when you hover over them now.

jburel:
Merging

PR #259, opened 2018-09-22
Promote next workshop

jburel:
Remove GBI
next stop Edinburgh

### DIFF
--- a/index.html
+++ b/index.html
@@ -8,7 +8,7 @@ title: Home
                 <div class="small-12 columns bg-image-text-container">
                     <h1 class="hero-main-header small-12">Next Workshop</h1>
                     <div align="center">
-                    <p align="left" class="small-10 small-offset-1 medium-6 medium-offset-3">18 October in Edinburgh (U.K.)<a href="http://eubias.org/NEUBIAS/training-schools/staff/ts9-edinburgh-2018/" target="_blank" class="hero-link">Neubias Schedule</a></p>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -7,10 +7,10 @@
             <div class="row homepage text-center">
                 <div class="small-12 columns bg-image-text-container">
                     <h1 class="hero-main-header small-12">Next Workshop</h1>
-                        <p class="small-10 small-offset-1 medium-6 medium-offset-3">19 - 21 September in Sydney (Australia)<a href="http://www.eurobioimaging.eu/content-page/international-training-courses-core-facility-staff" target="_blank" class="hero-link">GBI training Schedule</a></p>
+                    <div align="center">
```
mtbc: Probably also close this `div` somewhere.

index.html
```diff
@@ -7,10 +7,11 @@
...
+                    <p align="left" class="small-10 small-offset-1 medium-6 medium-offset-3">18 October in Edinburgh (U.K.)<a href="http://eubias.org/NEUBIAS/training-schools/staff/ts9-edinburgh-2018/" target="_blank" class="hero-link">Neubias Schedule</a></p>
```
will-moore: If you use class ```hero-link-inline``` instead of ```hero-link``` then the link will be on the same line as the text, which matches the later links and looks better. You might also need a space after ```(U.K.)```.

mtbc: Ah, that might help with some of https://trello.com/c/vFBFJAXC/156-workshops-formatting-issues.

jburel: If I remember correctly we did not want the schedule to be on the same line for the upcoming workshop
That's why ``hero-link-inline`` was not used.
Happy to change it if we think it is better

PR #263, opened 2018-10-08
Announcement fpbioimage 0.3.0

jburel:
Announcement

### DIFF
--- a/_posts/2018-10-09-fpbioimage-0-3-0.md
+++ b/_posts/2018-10-09-fpbioimage-0-3-0.md
@@ -1,0 +1,16 @@
+---
+layout: post
+title: Release of OMERO.FPBioimage 0.3.0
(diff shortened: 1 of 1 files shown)


snoopycrimecop:
Empty PR description. Please add a short summary of the PR scope and some testing instructions.

mtbc:
Staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/2018/10/09/fpbioimage-0-3-0.html.

mtbc:
Slightly strange that OMERO.FPBioimage 0.3.0 brings us Version 4 of FPBioimage but I assume that makes sense to somebody.

jburel:
It is the versioning using by FPBioimage, not under our control.
It is not a version our plugin that we can have as a major version

jburel:
Merging 
This will be tagged with 5.4.9

PR #269, opened 2018-11-15
Update explore dev/sysadmin examples

manics:
~This makes volumes, networks and internal container names more explicit. Also adds OMERO ports.~

Removes the embedded ansible and docker compose, links to external repos instead.


kennethgillen:
Looks nice at eff5a75 and it runs successfully.

joshmoore:
:+1: for the improvements as well. Also happy to see this file in a separate repo (where someone can more easily clone/wget it) but that raises the issue of how to keep them in sync. Options I've thought of:

 * Just download from github
 * Add as a submodule
 * Run `docker-compose config` to generate a variable-less version of something like `omero-test-infra`'s config

joshmoore:
With all the discussion around requirements today, I came back to https://github.com/ome/ansible-examples-omero/pull/12 . Though the name is not ideal, this docker-compose does look like it matches the `three-node` ansible playbook. Thoughts on using that as the location for this file and in this repo it gets downloaded?

manics:
Do you want to merge  ome/ansible-examples-omero#12 then so I can incorporate suggestions? I think we should keep the ansible and docker examples in separate directories because they're not directly comparable and they have different run-time requirements. Also a three-node OMERO.server ansible setup is very different from the docker deployment, in the former you're explicitly choosing to split your services across and therefore have to manage the orchestration and configuration, whereas Docker abstracts away the complexity of linking the three containers, and in production there's no such thing as a "single docker OMERO".

joshmoore:
Sounds like this needs a call. Several of your points I had the opposite suggestion from others.

joshmoore:
a-e-o 12 is merged. Happy to discuss unblocking this further.

manics:
Removed the embedded ansible and docker compose, link to external repos instead.

Screen Shot 2019-06-13 at 14 32 59-fullpage

joshmoore:
Looks great, @manics. Thanks.

The "Useful DevOps links" section needs updating. One link to the old "ansible-examples-omero" location. Perhaps we could change that to "Production ansible roles" and "Production docker images"?

Say the word if you want to fix that here or if you'd like to deploy this first.

manics:
Updated

Screen Shot 2019-06-17 at 14 44 30-fullpage

joshmoore:
:+1: for the general change (deployed at https://snoopycrimecop.github.io/www.openmicroscopy.org/explore/)

Long-term,  `dev-ops-links` seems oddly specific, but happy to get this out. Another longer-term question is where _else_ do we see linking to these things, since people tend to *not* find `/explore`?

sbesson:
Looks good. Noticed a few minor cases to update but I will tag and deploy immediately and open a separate PR for my changes.

sbesson:
Now deployed live at https://www.openmicroscopy.org/explore/

PR #271, opened 2018-11-19
iviewer announcement fix

jburel:
Add link

cc @manics


jburel:
I am going to redeploy so we can fix the announcement

PR #274, opened 2018-11-23
Fairsharing links

joshmoore:
Adding fairsharing links to https://www.openmicroscopy.org/about/ 👍 

screen shot 2018-11-26 at 12 09 58

Please see https://snoopycrimecop.github.io/www.openmicroscopy.org/about/#fairsharing for evaluation.

cc: @sbesson @jrswedlow @jburel


joshmoore:
Longer-term, this page likely needs to be broken up. I'd move most sections to their own sub-item under the About drop-down:
 * Consortium
 * Standards
 * Resources
 * (possibly) Funding, Goals

sbesson:
Having looked at the sitemap, I think this is also my preferred location on the website. At least that is the one location which references 3d party projects and how we integrate with them.

joshmoore:
Updated, now with #fairsharing anchor.

about/index.html
```diff
@@ -154,8 +154,41 @@ <h5 class="text-center">Places that publish image data with OMERO and Bio-Format
...
+        <hr>
+        <a name="standards"/>
+        <div id="standards" class="row small-up-2 medium-up-3">
+            <h5 class="text-center">
+                <a name="fairsharing"/>
+                <a target="_blank" href="https://fairsharing.org/collection/OME/">
+                OME-driven activities are collected in the
+                <a target="_blank" href="https://fairsharing.org/collection/OME/">OME Collection</a>
+                as a part of the
+                <a target="_blank" href="https://fairsharing.org/communities"/>Fairhsharing</a> community:
```
manics: typo: `Fairhsharing`

joshmoore: Ta.

jburel:
The OME collections link does not bring anything ome related

joshmoore:
> The OME collections link does not bring anything ome related

Weird. Thanks. Pushed.

jrswedlow:
Looks great.  I'd agree with @joshmoore  that we need sub-items. Linking to FAIRsharing doesn't provide good thumbs etc, which they might think about solving.  Lots of white space-- possible to get all 5 OME items in one horizontal line?  Other than that, is fine to push as  it is.

joshmoore:
I'm not seeing a quick way to get 5-wide, but I've fixed some of the whitespace (and updated the screenshot in the PR description). Follow-ons recorded in https://trello.com/c/mmJr2c2n/158-split-about-page

PR #277, opened 2019-01-07
Update Glencoe's team page link

emilroz:
This PR updates Glencoe's team page link.


joshmoore:
Thanks, @emilroz. Merging. @jburel / @sbesson : are there any other planned changes or shall I release just this?

jburel:
Going back to it. We will need to update the front page
i.e. plane, We can either do 2018->2019 transition or remove date from plane so we do not have to update it every year

PR #278, opened 2019-01-08
Next workshop

jburel:
* Add new workshop to the list
* Hide the year by changing opacity in case we want to replace 2018 by 2019

cc @joshmoore @sbesson


joshmoore:
New image is a slightly different size. Don't know if that will have any CSS impact. cc: @lunson

will-moore:
Looks OK. Year is removed from the banner.
Do we need to update the Events page to link to these?
Also, if users are interested by the front-page headlines, it would be nice if they could click on them, instead of having to know to go to Training / Events etc, but I guess that's a separate issue from this PR.

jburel:
We do not have event pages for the new events. I can't find a link for eastbio and nothing for singapore yet

jburel:
@joshmoore the updated image should now be the size of the old image

joshmoore:
@jburel : LGTM, though feel free to rebase away the unused 4KB and/or merge.

PR #287, opened 2019-04-03
Make users' workshop dates less ambiguous.

mtbc:
Easier for an international audience to read at a glance. Staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/events/omero-users-workshop-dundee-april-2019.html and https://snoopycrimecop.github.io/www.openmicroscopy.org/events/omero-users-workshop-dundee-june-2019.html.

### DIFF
--- a/events/omero-users-workshop-dundee-april-2019.html
+++ b/events/omero-users-workshop-dundee-april-2019.html
@@ -18,7 +18,7 @@ main-blurb: Two days workshop on how to use OMERO held in Dundee.
         </div>
         <div class="row">
             <div class="column">
-                <h2 class="event-day">Date: 08-09/04/2019</h2>
+                <h2 class="event-day">Date: 8th-9th April 2019</h2>
             </div>
         </div>
         <hr class="whitespace">
--- a/events/omero-users-workshop-dundee-june-2019.html
+++ b/events/omero-users-workshop-dundee-june-2019.html
@@ -18,7 +18,7 @@ main-blurb: Two days workshop on how to use OMERO held in Dundee.
         </div>
         <div class="row">
             <div class="column">
-                <h2 class="event-day">Date: 10-11/06/2019</h2>
+                <h2 class="event-day">Date: 10th-11th June 2019</h2>
             </div>
         </div>
         <hr class="whitespace">


events/omero-users-workshop-dundee-april-2019.html
```diff
@@ -18,7 +18,7 @@ <h2 class="event-day">Location: Dundee</h2>
         </div>
         <div class="row">
             <div class="column">
-                <h2 class="event-day">Date: 08-09/04/2019</h2>
+                <h2 class="event-day">Date: 8th-9th April 2019</h2>
```
sbesson: https://snoopycrimecop.github.io/www.openmicroscopy.org/events/ gives `April 08-09, 2019`. Worth unifying the date format ?

mtbc: I don't think so: the other events aren't unified and the context is different. (The index is a tiny note next to the calendar icon, this page is rather more verbose about the event.)

mtbc:
> or switching to ISO-8601

I do like to use ISO-8601 but in this case if I were going for a less-minimal change then I'd probably use a format that includes day of week: that's useful extra information for figuring meetings/conferences. That's probably a less urgent concern though.

mtbc:
I have been thinking about how most nicely to include day of week in date ranges. What would you think of something like,

> Date: Monday 8th to Tuesday 9th April 2019

Too verbose for this page?

No rush of course.

sbesson:
Sorry, I wasn't sure you were investigating date formats. I primarily merged this as my comment was not particularly a blocker and I wonder whether the training team wanted to redeploy the website before Monday's event /cc @jburel @pwalczysko 

Looking at date formats used by external training events, I like http://www.nibb.ac.jp/abis/gbi-abis_training?lang=en which is quite similar to your proposal.

mtbc:
Sure, no problem, made sense to merge this PR, can always revise at our leisure.

Maybe flip around to US ordering too then; I wasn't sure about that as most of our training is European anyway.

jburel:
We have tried to keep it not too long on that page 
For Date format, so far very limited session in US but I guess the format needs to match the rest of the website

PR #291, opened 2019-04-09
Typo

jburel:
fix typo in name
cc @sbesson


sbesson:
The French man agrees it is a typo :)

PR #295, opened 2019-05-27
support: point "Forum" to image.sc

joshmoore:
Currently using the somewhat empty "ome" tag. Alternatives include:

 - https://forum.image.sc/g/ome
 - https://forum.image.sc/g/ome/activity/posts
 - https://forum.image.sc/search? - with several or all tags like
   `expanded=true&q=tags%3Aomero%2Come%2Cbio-formats%2Come-tiff%2Come-xml`

### DIFF
--- a/support/index.html
+++ b/support/index.html
@@ -16,9 +16,9 @@ title: Support
                 <div class="card card-support">
                     <div class="card-section">
                         <i class="border-red icon-red fa fa-comments-o fa-2x"></i>
-                        <h4>OME Forum</h4>
-                        <p class="card-caption">Our forums allow the whole community to share their expertise and solutions.</p>
-                        <a href="https://www.openmicroscopy.org/community/" target="_blank" class="btn-blue button small">Visit the Forum</a>
+                        <h4>Forum</h4>
+                        <p class="card-caption">The Scientific Community Image Forum allows the whole community to share their expertise and solutions.</p>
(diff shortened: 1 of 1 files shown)


mtbc:
In the longer term: It's not much obvious to a new user that there are tags, which ones are relevant nor that this link is limiting to `ome` tag. So, maybe there's some thinking to do there but certainly no reason to block this PR.

sbesson:
Main feedback from my side is that the `ome` tag would work if we expect it to be able to represent a group of all our topics. In the absence of proper tag groups, I have a slighter preference for https://forum.image.sc/g/ome as the current landing page and update the description to point at the various tags/products that we are managing and watching.

joshmoore:
Alternatively, shall we point to www.openmicroscopy.org/forums which we can redirect more easily?

sbesson:
So define our internal redirect point via playbook and update as necessary? I would be happy with this solution (if we got things migrated I would assume `/community` would redirect as well in the long run)

joshmoore:
>  I would assume /community would redirect as well in the long run)

Yup.

Other options for the redirect:
 - /forums2
 - /new-forums
 - /image.sc
 - /community2
 - /imaging-community

joshmoore:
WFM

sbesson:
https://www.openmicroscopy.org/support/ now updated to point at the new forum.

PR #297, opened 2019-06-05
Update support page with image.sc forum instead of old forums and lists

will-moore:
As requested by @pwalczysko (```#training channel 22:33```) - updated the support page with the image.sc forum instead of old forums and mailing lists.

Do we want to link to archives, or will they be ported soon enough?

### DIFF
--- a/support/index.html
+++ b/support/index.html
@@ -18,8 +18,11 @@ title: Support
                         <i class="border-red icon-red fa fa-comments-o fa-2x"></i>
                         <h4>Image.sc Forum</h4>
                         <p class="card-caption">
-                            Please use "ome", "omero" or "bio-formats" tags on the Image.sc Forum when posting or
-                            searching.
+                            Please use the
+                            <a href="https://image.sc/tags/ome">"ome"</a>,
+                            <a href="https://image.sc/tags/omero">"omero"</a> or
+                            <a href="https://image.sc/tags/bio-formats">"bio-formats"</>
+                            tags on the Image.sc Forum when posting or searching.
                           </p>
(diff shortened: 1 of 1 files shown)


support/index.html
```diff
@@ -16,44 +16,25 @@ <h1>Support</h1>
...
                         <i class="border-red icon-red fa fa-comments-o fa-2x"></i>
-                        <h4>Forum</h4>
-                        <p class="card-caption">The Scientific Community Image Forum allows the whole community to share their expertise and solutions.</p>
-                        <a href="https://www.openmicroscopy.org/forums" target="_blank" class="btn-blue button small">Visit the Forum</a>
+                        <h4>Image.sc Forum</h4>
+                        <p class="card-caption">
+                            Please use "ome", "omero" or "bio-formats" tags on the Image.sc Forum when posting or
+                            searching.
+                          </p>
+                        <a href="https://forum.image.sc/" target="_blank" class="btn-blue button small">Visit the Forum</a>
```
joshmoore: :+1: for updating the name and pointing to tags, but the URL was specifically chosen so we can redirect in the future.

joshmoore:
> Do we want to link to archives, or will they be ported soon enough?

I wouldn't for today to reduce confusion. We can re-evaluate after this week.

cc: @pwalczysko

joshmoore:
Merging and releasing so this is in place for day 2 of ELMI. Happy to iterate if there are any issues.

Note: I'm unsure if there's a color policy on the icons other than ("roughly the same number of each"). We might want to place the rule in the HTML as a comment though. I certainly would have missed to update them. (See gh-298)

pwalczysko:
@will-moore Thank you very much for this quick action

PR #298, opened 2019-06-06
Drop ome-users and ome-devel

joshmoore:
see:
 * https://lists.openmicroscopy.org.uk/pipermail/ome-users/2019-June/007485.html
 * https://lists.openmicroscopy.org.uk/pipermail/ome-devel/2019-June/004377.html


joshmoore:
Closing in favor of gh-297

PR #311, opened 2019-08-29
Announce insight 5.5.5

jburel:
Same as changelog


mtbc:
Now staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/2019/08/29/insight-5-5-5.html.

PR #315, opened 2019-09-27
fix grammar typo

mtbc:
See diff.


mtbc:
Happy for you to decide the deployment schedule. Probably most people who were going to look at the page already did anyway. Thank you for asking though.

PR #331, opened 2020-01-09
Add sitemap

lunson:
Adds a sitemap using the plugin available here: https://github.com/kinnetica/jekyll-plugins. (This site map is not the same as https://www.openmicroscopy.org/site-map/ as it's specifically for Google to crawl.)

This is part of an effort to improve the SEO of the website and increase its website score, which as of this PR, is 80 (good but not great) and needs help specifically in the SEO area: https://website.grader.com/results/openmicroscopy.org

(I did a similar exercise for the GS website and our score has gone up from 79 to 97.)

Needs review - I'm having trouble including the blog posts from https://blog.openmicroscopy.org/  if you wish to include those here, but I think it would be okay for the blog to have its own separate crawl-able sitemap since it's a different website..

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -40,3 +40,10 @@ omefiles:
 
 omecmakesuperbuild:
  version: 0.5.0
+
+sitemap:
+    filename: "/sitemap.xml"
+    exclude:
+    include_posts:
+    change_frequency_name: "change_frequency"
+    priority_name: "priority"
--- a/_config.yml
+++ b/_config.yml
@@ -1,7 +1,7 @@
 name: OME  Website
 markdown: kramdown
 
-url: ''
+url: https://www.openmicroscopy.org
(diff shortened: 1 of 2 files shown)


joshmoore:
I'm getting a 404 at https://snoopycrimecop.github.io/www.openmicroscopy.org/sitemap.xml though I'm inclined to get this in and if it doesn't fail the build to see what happens on a production release.

_config.yml
```diff
@@ -1,7 +1,7 @@
 name: OME  Website
 markdown: kramdown
 
-url: ''
+url: https://www.openmicroscopy.org
```
joshmoore: This _may_ cause some issues with the snoopy deployment.

joshmoore:
After the build, http://www.openmicroscopy.org/sitemap.xml is 404.

PR #333, opened 2020-01-17
add sitemap - part 2

lunson:
Trying to fix what might've gone wrong with https://github.com/ome/www.openmicroscopy.org/pull/331

The sitemap appears for me locally (but then again, it appeared for me locally for PR #331 too), so hopefully this works.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,7 +1,7 @@
 name: OME  Website
 markdown: kramdown
 
-url: https://www.openmicroscopy.org
+url: ''
 repository: openmicroscopy/www.openmicroscopy.org
 baseurl: /www.openmicroscopy.org
 


joshmoore:
Deploying as 2020-01-17

joshmoore:
Deploying as 2020-01-17

joshmoore:
Unfortunately, no change @lunson

lunson:
Sorry. Tabling for now. I don't know enough about the build process (or builds in general) to properly set up the config file. Stick helped me with the GS version and I mistakenly thought copying what was done for that site would result in the same output.

joshmoore:
No worries. I'll see if @stick can help me understand the difference.

PR #335, opened 2020-01-17
Fix link to insight

jburel:
* It is now a jar and no longer a zip
cc @pwalczysko


joshmoore:
Deploying as 2020-01-17

PR #336, opened 2020-01-17
Fix site.omero.build number.

will-moore:
The b138 link is 404 just now, but b136 works

Josh just fixed this in-place so that the server download link is no-longer broken.


joshmoore:
Thanks for catching my mistake, @will-moore!

PR #340, opened 2020-02-05
Add new sitemap

lunson:
Adds a sitemap that actually works :)


joshmoore:
:) :+1:

chris-allan:
gh-pages will be using jekyll-sitemap 1.4.0 as of this writing (https://pages.github.com/versions/) but the `jekyll/builder:pages` Docker container currently used by Travis is sitting on jekyll-sitemap 1.2.0. We haven't seen issues on our side with those version differences but it's something to watch out for when comparing https://snoopycrimecop.github.io/www.openmicroscopy.org/sitemap.xml with the eventual https://www.openmicroscopy.org/sitemap.xml.

/cc @manics

manics:
https://hub.docker.com/r/jekyll/jekyll
- links to https://github.com/jekyll/docker/tree/master/README.md
- which redirects to https://github.com/envygeeks/jekyll-docker/blob/master/README.md

Maybe someone forgot to update the image?

chris-allan:
Certainly possible. I think the crux is the installation of the version 201 of the github-pages gem rather than the 204; which is the latest any only released a couple weeks ago.

It's probably suggested to use bundler to set versions correctly but the amount of hassle I had trying to get that to function the way I wanted with the `jekyll/builder:pages` container was immense. If you want to have a bash by all means but it's a multi-day rabbit hole I expect.

joshmoore:
We currently have nothing. Propose to roll this into the next website update and then we go from there?

cc: @jburel

chris-allan:
@joshmoore: Absolutely. :+1:

jburel:
Having a site map will be great.
Let's include it in next release

lunson:
@joshmoore Please ping me when it's deployed :) I have to add it to Google Search Console afterward. Thank you!

joshmoore:
https://www.openmicroscopy.org/sitemap.xml is deployed.

lunson:
The domain is missing from all the URLs in the sitemap, so Google is saying it's invalid. It works on my local, so not sure how to fix on the build side. (Sorry.)

joshmoore:
My best guess is that it's going to require either changing or adding a variable, like `url`: https://github.com/ome/www.openmicroscopy.org/pull/340/files#diff-aeb42283af8ef8e9da40ededd3ae2ab2L4

chris-allan:
That's what we have on the glencoesoftware.com side so makes sense here.

@lunson: Try it with a setting `url: https://www.openmicroscopy.org` and we'll go from there.

PR #341, opened 2020-02-06
Add Susanne to contributors

joshmoore:
With thanks for the recent MDE work.

cc: @sukunis @jburel @dominikl


jburel:
Looks good.

PR #337, opened 2020-01-24
Store repeating data in YAML

lunson:
Taking advantage of YAML front matter and allowing it to create repeating divs for some of the website's more popular pages (according to Analytics): Home, OMERO, About. This will hopefully make it easier to make future updates to any text, and should reduce copy/paste errors. It certainly makes things easier for me to restyle repeating elements if need be :)

(More pages to come in future PRs...)

PR #346, opened 2020-03-06
Migrate alumni content into YAML format

sbesson:
Alongside the new format proposed in #337, this also migrates the content of the past team members list into YAML format /cc @lunson 

Staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/teams/


jburel:
Looks good
Merging so the page will be updated when we deploy next

PR #370, opened 2020-05-20
Reorder ome2020 schedule

will-moore:
This is on top of Liza's PR: https://github.com/ome/www.openmicroscopy.org/pull/369

Try to clarify the pre-recorded videos before discussions on day 1 and 2

PR #361, opened 2020-05-13
Add videos layout

lunson:
This is the proposed videos layout I mention in the May-13 email to @pwalczysko:
- It's more visually engaging that just using links
- It's mobile friendly
- Users can watch the video here without leaving the page

There was talk in one of the weekly OME meetings to perhaps use the same embedding method Glencoe Software used for their webinars: https://www.glencoesoftware.com/media/webinars/ so I included a sample Glencoe video here.

Someone had also mentioned perhaps using YouTube, so I included a sample (worth watching, I might add 😄) of that also.

Not ready for merge until we have the actual speaker videos, but I want to leave this here so you have something to work with in case those videos arrive after I've already left. (The `program_agenda.html` `include` file could probably use some improvement by someone who is more adept at `if` /`then` and `forloop` statements, as that is not my strength!)

Happy to take suggestions for improvement!

If this layout doesn't work for you guys, what was originally proposed should be just fine.

PR #372, opened 2020-05-20
Add real videos

pwalczysko:
cc @will-moore adding here two videos available at the moment, for ease of checking the layout with real content.

Feel free to cherry-pick the commits, I will close the PR then, or, we might decide otherwise later.


pwalczysko:
@will-moore I wonder why do the videos in the layout suggested in https://github.com/ome/www.openmicroscopy.org/pull/370  (this PR is on top of the https://github.com/ome/www.openmicroscopy.org/pull/370) are appearing black until you click on them to play.
Whereas in the layout which was in https://github.com/ome/www.openmicroscopy.org/pull/361, the videos were showing an image representing the first frame. I do not think this is because of the videos themselves. See screenshot below for https://github.com/ome/www.openmicroscopy.org/pull/361 and compare with https://snoopycrimecop.github.io/www.openmicroscopy.org/events/ome-community-meeting-2020/day1/ for the present, black one.

Screen Shot 2020-05-20 at 20 56 02

PR #391, opened 2020-06-16
htmlproofer: reenable external linkchecking for 404

sbesson:
The `htmlproofer` step had been disabled for all external link. This PR selectively re-enables it at least to check for 4xx errors (except 429 which are typically thrown by GitHub downloads URLs)

A few broken links are also fixed or removed when no alternate solution was found.

PR #394, opened 2020-07-07
Update for Bio-Formats 6.5.1

dgault:
Add release announcement and bump for Bio-Formats 6.5.1
See https://trello.com/c/G1ptHPOM/20-open-pr-against-wwwopenmicroscopyorg

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -15,7 +15,7 @@ paginate_path: "/announcements/page:num/"
 future: true
 
 bf:
- version: 6.5.0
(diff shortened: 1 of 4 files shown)


sbesson:
The minor changes from https://github.com/ome/bio-formats-documentation/pull/169#pullrequestreview-444200971 will need to be ported here as well.

The broken links are a side-effect of  #391 which re-enabled the linkchecking. As soon as the docs/downloads are publicly available, this should be ready for deployment.

PR #399, opened 2020-07-17
roll back header

jburel:
Remove the community banner
I have opted to hide the text so it is easier to switch back when we organise another meeting

cc @sbesson

### DIFF
--- a/index.html
+++ b/index.html
@@ -57,6 +57,7 @@ many_thanks:
         <!-- begin standard header to be hidden when event like community meeting is announced -->
         <header class="marketing-hero" id="bg-image-main">
             <div class="row homepage text-center">
+              <div class="small-12 columns bg-image-text-container">
               <h1 class="hide-for-small-only hero-main-header small-12">The Open Microscopy Environment</h1>
               <h1 class="show-for-small-only hero-main-header"><img src="{{ site.baseurl }}/img/logos/ome-white.svg" alt="OME logo" class="logo-hero" /></h1>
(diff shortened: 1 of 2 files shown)


sbesson:
ONly other thought is: is it worth promoting the latest OMERO release until the next announcement?

will-moore:
It doesn't look quite how it used to (screenshot is from a while back) - just trying to work out why...

Screenshot 2020-07-17 at 12 27 24

index.html
```diff
@@ -53,7 +53,23 @@
         </section>
         <!-- end banner announcement -->
 
+        
+        <!-- begin standard header to be hidden when event like community meeting is announced -->
+        <header class="marketing-hero" id="bg-image-main">
+            <div class="row homepage text-center">
```
will-moore, line 59: You need to add class "bg-image-text-container" here.
Also the -begin banner announcement- section above needs to be inside the commented-out section below.
It was added for the OME meeting with "register now". Then the "register now" was removed when the registration closed but I didn't remove the banner itself since the page would look wrong without the red line.

jburel:
@will-moore better now?

will-moore:
Looks good 👍 

Screenshot 2020-07-17 at 13 25 21

NB: I like the link to latest release, but in general I would try to limit the number of links here.
3 is OK, but 1 or 2 is better since I think that this isn't supposed to be a list of options but really just the top 1 or 2 most obvious ones?

will-moore:
Good to merge.

jburel:
Thanks @will-moore for the review

PR #405, opened 2020-08-06
Exchange icons

pwalczysko:
Follow-up of the discussion this am.
This is on top of https://github.com/ome/www.openmicroscopy.org/pull/401 -  but I opted to open as a separate PR - happy to be told otherwise.

Note that there were other places which were also using the old style icons, such as Downloads of both OMERO and BF.

I am opening this as a separate PR as it is not as small a change - and probably we should have a look at the site whether or not still some icons persist.

Edit:  Note that practically only the last 2 commits are really new for this PR. The rest is coming from https://github.com/ome/www.openmicroscopy.org/pull/401

Built on https://snoopycrimecop.github.io/www.openmicroscopy.org/omero/features/analyze/  - but also check other pages, like downloads and the New features etc.

cc @jrswedlow @jburel @joshmoore @sbesson


pwalczysko:
closing this, the review is happening in https://github.com/ome/www.openmicroscopy.org/pull/401 and I cherry-picked the commits from here to https://github.com/ome/www.openmicroscopy.org/pull/401

Issue #414, opened 2020-10-13
Move static content to GitHub pages


manics:
https://docs.github.com/en/free-pro-team@latest/github/working-with-github-pages/about-github-pages

1. Since this will be a top level site `https://www.openmicroscopy.org/` it has to be in a repo called `<USER-OR-ORG>.github.io`, so either `ome/ome.github.io` or use new GitHub user e.g. `ome-www/ome-www.github.io`.
2. Add a `CNAME` file to this repo containing `www.openmicroscopy.org`
3. Add a DNS CNAME `www.openmicroscopy.org ➔ <USER-OR-ORG>.github.io`
4. Edit the `baseurl` in `_config.yml`

sbesson:
Adding a few thoughts, potentially to be captured as separate issues:

- why is the new repo a requirement? we have ome/ome-help and ome/omero-figure successfully deploying to help.openmicroscopy.org and figure.openmicroscopy.org 
- in addition to the static website, the content the playbooks under https://github.com/ome/prod-playbooks/tree/master/www will need to be reviewed and at least the critical parts added to the config
- from a quick glance, the most important bit missing are the XSD schemas hosted under `www.openmicroscopy.org/Schemas` previously proxied from another server. Some in-progress work and thoughts are captured in https://github.com/ome/www.openmicroscopy.org/pull/286#issuecomment-707581528 but this requires some more input 
- from previous experience with help/figure, one downside of deploying with GitHub pages using a custom CNAME is that it complicates the staging of an integration branch. Alternatively would there be a mechanism to stage per-PR builds?

manics:
1. I was going by what's implied by the docs, but if you've got a working configuration that's great!
4. I think the tradeoff between easier production deployment and resilience is worth the additional configuration required on the CI side. One workaround is to have a permanently open PR that's changes the configuration of the build site.

PR #417, opened 2020-10-14
www.openmicroscopy.org CNAME for GH page hosting

manics:
Closes https://github.com/ome/www.openmicroscopy.org/issues/414

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,0 +1,1 @@
+www.openmicroscopy.org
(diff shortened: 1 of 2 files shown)


joshmoore:
A follow-up might be updating the banner to say that some links might be broken?

sbesson:
OK switched the upstream CNAME record. Let's try this. Agreed with https://github.com/ome/www.openmicroscopy.org/pull/417#issuecomment-708271876 as an immediate follow-up

PR #422, opened 2020-10-15
Fix broken navbar html

manics:
The current HTML is broken


joshmoore:
Thanks, @manics. I see no reason not to get this in.

PR #423, opened 2020-10-15
Revert "www.openmicroscopy.org CNAME for GH page hosting"

sbesson:
Reverts ome/www.openmicroscopy.org#417

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,1 +1,0 @@
-www.openmicroscopy.org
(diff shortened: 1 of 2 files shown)


sbesson:
Note the plan here is very much to re-revert this change and have the complete website (including the redirects) managed via GitHub pages.

As more investigation is required to fix the remaining links, this PR will temporarily allow to stage the state of the `master` branch  at www.openmicroscopy.org and allow testing/validation.

PR #426, opened 2020-11-25
Add deploy script to replace ansible playbook

manics:
Not yet tested in production.

To test locally override the parent directory. Optionally pass a version from https://github.com/ome/www.openmicroscopy.org/releases, default `latest`
```
./deploy.py --parent /tmp/my/test/directory --version 2020.10.01-1
./deploy.py --parent /tmp/my/test/directory
...
```

If this works as intended `deploy.py` could be run in a cronjob to automatically deploy the latest tag.

### DIFF
--- a/_deploy/deploy.py
+++ b/_deploy/deploy.py
@@ -1,0 +1,71 @@
+#!/usr/bin/env python
+import argparse
+import json
+import os
+import sys
+import tarfile
+# Python 2 and 3 compatible so this can be run on the RHEL7 without a
+# virtualenv
+try:
+    from urllib.request import urlopen
+except ImportError:
+    from urllib2 import urlopen
+
+
+def main(version, parent):
+    # https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#releases
+    # No paging, assume no-one will install a really old version
+    r = urlopen('https://api.github.com/repos/ome/www.openmicroscopy.org/releases')
+    assert r.code == 200
+    releases = json.load(r)
+
+    if version == 'latest':
+        release = releases[0]
+    else:
+        release = None
+        for check in releases:
+            if check['tag_name'] == version:
+                release = check
+                break
+        if release is None:
+            print('Failed to find release {}'.format(version))
+
+    tag = release['tag_name']
+
+    dst = os.path.join(parent, tag)
+    sym = os.path.join(parent, 'html')
+
+    if os.path.exists(dst):
+        print('{} already exists, not downloading'.format(dst))
+    else:
+        www_assets = [a for a in release['assets'] if a['name'] == 'www.openmicroscopy.org.tar.gz']
+        assert len(www_assets) == 1, 'Expected one asset named www.openmicroscopy.org.tar.gz'
+        url = www_assets[0]['browser_download_url']
+
+        h = urlopen(url)
+        thetarfile = tarfile.open(fileobj=h, mode="r|gz")
+        thetarfile.extractall(path=dst)
+        h.close()
+        print('Extracted {} to {}'.format(url, dst))
+
+    if os.path.exists(sym):
+        assert os.path.islink(sym), '{} is not a symlink'.format(sym)
+        if os.readlink(sym) == dst:
+            print('{} already points to {}, no changes made'.format(dst, sym))
+            sys.exit(0)
+        print(os.readlink(sym))
+        os.remove(sym)
+    os.symlink(dst, sym)
+    print('Symlinked {} to {}'.format(dst, sym))
+    sys.exit(0)
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '--parentdir', default='/var/www/www.openmicroscopy.org',
+        help='Web-server directory for www.openmicroscopy.org')
+    parser.add_argument('--version', default='latest',
+        help='Release to download')
+    args = parser.parse_args()
+    main(args.version, args.parentdir)
--- a/_deploy/deploy.py
+++ b/_deploy/deploy.py
@@ -4,8 +4,7 @@ import json
 import os
 import sys
 import tarfile
-# Python 2 and 3 compatible so this can be run on the RHEL7 without a
-# virtualenv
+# Python 2 and 3 compatible so this can be run on RHEL7 without a virtualenv
 try:
     from urllib.request import urlopen
 except ImportError:
@@ -29,6 +28,7 @@ def main(version, parent):
                 break
         if release is None:
             print('Failed to find release {}'.format(version))
+            sys.exit(1)
 
     tag = release['tag_name']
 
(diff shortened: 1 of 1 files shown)


joshmoore:
@manics : I pushed https://github.com/ome/www.openmicroscopy.org/pull/426/commits/d3c6e74360a923d96341ae8c3d2af51f7420d437 to allow testing.

For everyone else, cf:
```
Would download /var/www/www.openmicroscopy.org/2020.11.26
Would remove symlink: /var/www/www.openmicroscopy.org/html (target=/var/www/www.openmicroscopy.org/2020.10.07/)
Would symlink /var/www/www.openmicroscopy.org/2020.11.26 to /var/www/www.openmicroscopy.org/html
```

_deploy/deploy.py
```diff
@@ -0,0 +1,85 @@
...
+        # Mutator
+        os.symlink(dst, sym)
+        print('Symlinked {} to {}'.format(dst, sym))
+    sys.exit(0)
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    xor = parser.add_mutually_exclusive_group()
+    xor.add_argument('-n','--dry-run', action='store_true', default=True)
```
joshmoore, line 77: Would be nice if this were a standard API on ArgumentParser....

joshmoore:
Then with `--force`, 

```
[jamoore@ome-www www]$ sudo _deploy/deploy.py  -f
Extracted https://github.com/ome/www.openmicroscopy.org/releases/download/2020.11.26/www.openmicroscopy.org.tar.gz to /var/www/www.openmicroscopy.org/2020.11.26
/var/www/www.openmicroscopy.org/2020.10.07/
Symlinked /var/www/www.openmicroscopy.org/2020.11.26 to /var/www/www.openmicroscopy.org/html
```

https://www.openmicroscopy.org/events/ is updated :+1: cc: @pwalczysko 

NB:
```
[jamoore@ome-www www]$ du -sh /var/www/www.openmicroscopy.org/
1.9G	/var/www/www.openmicroscopy.org/
```

sbesson:
A few additional comments:

- if absolute minimal downtime was a requirement, the same symlink swap strategy as the one used in https://github.com/ome/scc/blob/master/scc/deploy.py could be implemented.
- slightly confused about why this script is in this repository, it seems that you need to clone/download www, run it and then re-download another copy. Should this script and the optional cron configuration be managed via https://github.com/ome/prod-playbooks/blob/master/www/playbook.yml eventually?
- is this aim to completely replacing the `deploy_archive` step in https://github.com/ome/prod-playbooks/blob/master/www/www-static.yml#L12. Main concern about the dual deployment strategy is that playbook could potentially override the website updated by this script?

joshmoore:
> * slightly confused about why this script is in this repository,

I had a similar thought. I could certainly see this logic living somewhere general even at the cost of needing to specify:

```
(scc deploygithub-release-deployer) ome/www.openmicroscopy.org /var/www/www.openmicroscopy.org/html
```

Note: there are several "github-release-deployer"-style tools on GitHub.

_deploy/deploy.py
manics: I don't think we need both flags, just one is fine. If you prefer to be safe then require `--force` and mention the default is to only display changes. If this is going to be run automatically then requiring the `--force` argument isn't a problem.

manics:
There's no single best place for this script so I opened it here since this is what it's related to and it seems a good place for the discussion. If someone wanta to revive the scc deploy behaviour I can close this, or this script could be moved to prod-playbooks, or a standalone repo?

_deploy/deploy.py
joshmoore: I find having both more posix-y.

pwalczysko:
I understand that the target server (the production ome-www server in this case) is hard-coded in the script ?
Even with a ``--dry-run`` option added, I still would think we should clarify:
What is the reason for dropping the step of deploying first on ``ome-www-dev`` ?
When something goes wrong in the sense that the release artifact itself, although deploying nicely (and thus ``--dry-run`` would pass) is somehow not sound, the first place we would find that out (if the ``ome-www-dev`` step is skipped) would be the direct live production op...py.org site ?

pwalczysko:
What is the purpose of the ``dry-run ? What warning should it give me ? See below the output with and without dry-run

```
(env562) pwalczysko@ls31618~/Work/www.openmicroscopy.org/_deploy$ python deploy.py --parent /tmp/my/test/directory --version 2020.10.01-1 --dry-run
Would download /tmp/my/test/directory/2020.10.01-1
Would symlink /tmp/my/test/directory/2020.10.01-1 to /tmp/my/test/directory/html
(env562) pwalczysko@ls31618~/Work/www.openmicroscopy.org/_deploy$ python deploy.py --parent /tmp/my/test/directory --version 2020.10.01-1
Would download /tmp/my/test/directory/2020.10.01-1
Would symlink /tmp/my/test/directory/2020.10.01-1 to /tmp/my/test/directory/html
(env562) pwalczysko@ls31618~/Work/www.openmicroscopy.org/_deploy$ 
```

pwalczysko:
Considering the output in https://github.com/ome/www.openmicroscopy.org/pull/426#issuecomment-739857592 and the remark in https://github.com/ome/www.openmicroscopy.org/pull/426#issuecomment-739854299 I have the feeling that the script has not been tested and is not really testable (unless used in an earnest, wet run) as things stand...

manics:
> What is the purpose of the dry-run ? What warning should it give me ? See below the output with and without dry-run

That's one of the reasons I suggested not having both `--dry-run` and `--force` in https://github.com/ome/www.openmicroscopy.org/pull/426#discussion_r531101533 since it causes confusion like this- the default is `--dry-run`.

pwalczysko:
Tested the script on ``ome-www-dev``

```
[pwalczysko@ome-www-dev _deploy]$ sudo python ./deploy.py --version 2020.10.01-1 -f
/var/www/www.openmicroscopy.org/2020.10.01-1 already exists, not downloading
/var/www/www.openmicroscopy.org/2020.12.07/
Symlinked /var/www/www.openmicroscopy.org/2020.10.01-1 to /var/www/www.openmicroscopy.org/html
[pwalczysko@ome-www-dev _deploy]$ sudo python ./deploy.py -f
/var/www/www.openmicroscopy.org/2020.12.07 already exists, not downloading
/var/www/www.openmicroscopy.org/2020.10.01-1
Symlinked /var/www/www.openmicroscopy.org/2020.12.07 to /var/www/www.openmicroscopy.org/html
[pwalczysko@ome-www-dev _deploy]$ 
```

It worked.

joshmoore:
:+1: Thanks, @pwalczysko. I hadn't thought about it, but @sbesson pointed out that the primary downside of using the script as it stands is that running the playbook will revert those changes. i.e. we will need to make some decisions and changes before moving forward with this.

manics:
Moved to https://github.com/ome/prod-playbooks/pull/299 (thanks @sbesson)

Issue #348, opened 2020-03-10
Add OMERO to licensing page


manics:
It's currently missing from https://www.openmicroscopy.org/licensing/

pwalczysko:
Happy to format it and open a PR, but what content should be in the blurb ? cc @manics @sbesson

joshmoore:
Much the same as the Bio-Formats section but without the sentences about readers ("some components which provide reader and writer implementations for open file formats, are released under a more permissive BSD-2 license which enables non-GPL third party software. For a complete list of which file formats are included in the BSD license, see the BSD column of the supported formats table. ")

pwalczysko:
Opened https://github.com/ome/www.openmicroscopy.org/pull/431

manics:
Closed by https://github.com/ome/www.openmicroscopy.org/pull/431

PR #431, opened 2020-12-07
Add OMERO to licensing

pwalczysko:
See https://github.com/ome/www.openmicroscopy.org/issues/348#issuecomment-740020699 for motivation

Adding OMERO to licensing page with the text suggested in https://github.com/ome/www.openmicroscopy.org/issues/348#issuecomment-740020699

cc @manics @joshmoore

### DIFF
--- a/licensing/index.html
+++ b/licensing/index.html
@@ -36,6 +36,17 @@ meta_description: OME software is open source and freely available through vario
 
         <!-- begin -->
         <hr class="whitespace">
+        <div class="row column text-center">
+            <h2>OMERO</h2>
+        </div>
+        <hr>
+        <div class="row column">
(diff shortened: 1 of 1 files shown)


joshmoore:
Screen Shot 2020-12-08 at 10 55 50

joshmoore:
I think this is a fine start. I imagine we could remove a bit of the redundancy by separating it down into licenses:

 - GPL
   - Bio-Formats
   - OMERO
 - BSD
   - Bio-Formats (BSD)
   - OME-Files

etc. but I don't propose to do that kind of overhaul here.

Any other thoughts, @sbesson @jburel @jrswedlow @chris-allan ?

sbesson:
Agreed this might need reworking to express the granularity of the licensing between the different components of each product. I don't have an immediate suggestion so let's start with clarifying the OMERO expectations on the website as it stands and then coming back to this in a second step. Thanks @pwalczysko

sbesson:
Now deployed at https://www.openmicroscopy.org/licensing/

PR #432, opened 2020-12-08
Checkout submodules for build

joshmoore:
Note that https://www.openmicroscopy.org/minutes/ currently is empty as opposed to https://snoopycrimecop.github.io/www.openmicroscopy.org/minutes/

Thanks to June for pointing it out. Propose for immediate tagging & re-release.

PR #433, opened 2020-12-08
Add checks for file existence

sbesson:
As a follow-up of #432, this adds minimal checks to the `build.sh` script to ensure at least one reference file from each submodule is in the build


pwalczysko:
failed checks

sbesson:
The check suite is now green using the merge commit built from this PR and `origin/master` with #432 included.
For comparison, the build from d7debcc alone (without #432) failed as expected due to the absence of the submodules - see https://github.com/sbesson/www.openmicroscopy.org/runs/1518467586?check_suite_focus=true

joshmoore (approved):
LGTM

PR #435, opened 2020-12-14
Update for Bio-Formats 6.6.0

dgault:
Add release announcement and bump for Bio-Formats 6.6.0


sbesson (commented):
Looks good. The failing builds require the docs/downloads to be updated.

PR #439, opened 2021-01-06
Twitter links

sbesson:
For some still unclear reasons, the Twitter handle URLs started failing `htmlproofer` with a 400 error code. 

As this is preventing website releases, this is updating the `htmlproofer` command to ignore the Twitter URLs. I can add a follow-up issue to review.


joshmoore:
Perhaps due to "...does not use Twitter's API, thus, I'm limited to 150 requests per hour, per IP address. " ([stackoverflow](https://stackoverflow.com/questions/13017025/twitter-widget-returning-400-bad-request-error))

A compromise in a potential follow-up might be to exclude these URLs that we know to be safe (and have substantial overhead for loading) and instead only check explicit tweets (if we even have any).

Otherwise, :+1:

PR #442, opened 2021-02-01
Move Simon to the former OME members section

sbesson:
Simon moved to a new position. Also updates the broken URL for Ilan Davis group


joshmoore:
Barring the thought we could also include twitter ;) LGTM.

sbesson:
Yes Twitter would be nice but I'll separate from this. Can be captured as an issue

PR #456, opened 2021-03-18
Contributors: additions and data files

sbesson:
This PR adds recent OMERO contributors to the contributors page.

a19488f is a proposal to format contributor sources as [Jekyll data files](https://jekyllrb.com/docs/datafiles/). YAML was chosen but CSV would be an option as well. The HTML template is simply looping over the dictionaries and generating the various entries including the new optional GitHub link.

9869412 adds various recent contributors to the decoupled OMERO repositories, primarily omero-py and omero-web.


sbesson:
Staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/contributors/

joshmoore (approved):
Screen Shot 2021-03-22 at 12 16 28


 * https://snoopycrimecop.github.io/www.openmicroscopy.org/contributors/ looks good (though I could see spicing it up later with jekyll-avatar ;) )
 * :100: for moving more to data files. _Much_ easier to maintain (though as mentioned elsewhere, I'd eventually propose going to one file per person. To be tested on /teams first)
 * The contributor additions all look fine, but I can't say if there are any that have been missed.
 * Somewhere I had a bot or GH action which automatically updated this type of list. That may be a route down the road.

:+1:

sbesson:
> The contributor additions all look fine, but I can't say if there are any that have been missed.

I have been through the most recent additions mentioned in the CHANGELOG of omero-py & omero-web mostly but we might need an more thorough audit.

> Somewhere I had a bot or GH action which automatically updated this type of list. That may be a route down the road. 

That's something I'll consider in my upcoming investigation on the CLA.

:+1: for follow-up improvements to contributors, teams, projects pages.

sbesson:
Now deployed at http://www.openmicroscopy.org/contributors/

PR #461, opened 2021-03-31
OME2021 community meeting

sbesson:
- 4a7d157  copies/duplicates the templates for OME2020 and creates a new OME2021 event
- 6f88499 creates a JSON data file following https://schema.org/Event to model the event and updates the layouts/templates to consume it

### DIFF
--- a/_data/ome2021.json
+++ b/_data/ome2021.json
@@ -1,0 +1,340 @@
+{
+  "@id": "https://www.openmicroscopy.org/events/ome-community-meeting-2021/",
+  "@context": "https://schema.org",
+  "@type": "Event",
+  "eventAttendanceMode": "https://schema.org/OnlineEventAttendanceMode",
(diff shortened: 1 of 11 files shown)


joshmoore:
I think this is great! As long as we're not linking to the new page explicitly, I'd be for getting it in so that we can work on it in parallel.

sbesson:
No objections to getting this in sooner rather than later and distribute follow-up work. From my side at least, I plan to look at the following next:

- get the website re-indexed and review the rich search results
- add the videos to the event - I have some preliminary work using https://schema.org/workFeatured
- expand the introduction page and the agenda as we progress

sbesson:
As the first round of invitations are about to be sent, having a first version of a public page detailing the program (with a clear in-progress) seems to be the best consensus. Merging and deploying.

PR #466, opened 2021-04-09
Add cli-duplicate to Apps

pwalczysko:
Adding an obviously missing app to the list.

There are several others which could be added

  - omero-napari
  - OMERO.dropbox (not a CLI neither a Web app though, so would need a new paragraph, maybe called Server apps ?)
  - omero-signup
  - omero-cli-zarr
  - omero-roi
  - ...
 
Not sure about any of these, and so trying to go just with the duplicate in this PR - happy to be told otherwise.

### DIFF
--- a/omero/apps/index.html
+++ b/omero/apps/index.html
@@ -162,3 +162,17 @@ meta_description: Learn how to extend the functionality of OMERO with web or com
                 </a>
             </div>
         </div>
+        <div class="row">
+            <div class="medium-1 medium-offset-2 columns">
+                <i class="icon-fa-blue fa fa-share-alt fa-2x"></i>
+            </div>
+            <div class="omero-key-feature-container medium-7 columns end">
+                <h4>OMERO-<span class="lowercase">cli-duplicate</span></h4>
+                <p>
+                    Create data duplicates.
+                </p>
+                <a class="button hollow tiny" href="https://pypi.org/project/omero-cli-duplicate/" target="_blank">
+                    <span class="lowercase">omero-cli-duplicate</span>
+                </a>
+            </div>
+        </div>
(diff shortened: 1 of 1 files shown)


omero/apps/index.html
```diff
@@ -162,3 +162,17 @@ <h4>OMERO-<span class="lowercase">metadata</span></h4>
...
             </div>
         </div>
+        <div class="row">
+            <div class="medium-1 medium-offset-2 columns">
+                <i class="icon-fa-blue fa fa-share-alt fa-2x"></i>
+            </div>
+            <div class="omero-key-feature-container medium-7 columns end">
+                <h4>OMERO-<span class="lowercase">cli-duplicate</span></h4>
+                <p>
+                    Create data duplicates.
```
joshmoore, line 172: "Duplicate Objects in OMERO, without duplicating binary data." ? (from the repo)

joshmoore:
> There are several others which could be added

Yeah, this is going to be a continual question. At the moment, the list is quite curated. Perhaps dependent on what we have guides for? Somewhere there should also be a complete catalogue. Unclear if that should be here as well.

omero/apps/index.html
pwalczysko: fixed in https://github.com/ome/www.openmicroscopy.org/pull/466/commits/15b5e3a46fc85993add921d6531a8c3790fc2771

pwalczysko:
> > There are several others which could be added
> 
> Yeah, this is going to be a continual question. At the moment, the list is quite curated. Perhaps dependent on what we have guides for? Somewhere there should also be a complete catalogue. Unclear if that should be here as well.

Thank you @joshmoore . I like the suggestion "if there is a guide, we list it here". The only one with a guide which is not here is OMERO.dropbox. For that one, I have a  question "How to highlight the app in the sections listed here ? There is no "server apps" section, so add a new section called "server apps" ?  Is that accurate ?

will-moore:
Looks good. 👍

pwalczysko:
Thanks @will-moore . My vote would be for this to merge, as https://github.com/ome/www.openmicroscopy.org/pull/466#issuecomment-817645515 needs a further discussion/thinking, whereas the present change here is ready to go. Happy to draft a suggestion in a new PR cc @joshmoore

PR #467, opened 2021-04-13
Add ORCID IDs to team JSON and webpage

sbesson:
As ORCIDs are becoming an authoritative source of identification especially in the academic world, this contains the necessary changes to:

- add the ORCID IDs of the team to the schema.org JSON+LD as `@id` similarly to other projects - https://www.researchobject.org/ro-crate/1.1/contextual-entities.html#people
- add an inline ORCID iD (see https://info.orcid.org/brand-guidelines/#orcid-logos-and-icons) to the teams page

### DIFF
--- a/_data/swedlowlab.json
+++ b/_data/swedlowlab.json
@@ -8,7 +8,8 @@
       "givenName": "Jason",
       "familyName": "Swedlow",
       "jobTitle": "Co-Founder",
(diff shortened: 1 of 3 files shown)


sbesson:
Now deployed at https://snoopycrimecop.github.io/www.openmicroscopy.org/teams/

joshmoore:
Love the icons. (Could imagine having GH ones, too) But the bios are no longer opening for me.

sbesson:
> But the bios are no longer opening for me.

Ah you're right. Or rather it looks like for those who have ORCID IDs associated, clicking on the picture rather than `Read Bio` now loads the icon. I'll take a look and relist.

joshmoore:
As mentioned elsewhere, as we get in to adding more external (social media) links, I could see moving to a single `sameAs` block at the risk of slightly complicating the consuming code:

```
{
  "@context":"http:\/\/schema.org\/",
  "@type":"Person",
  "sameAs":[
    "https:\/\/twitter.com\/foo",
    "https:\/\/github.com\/bar"
  ]
```

sbesson:
I think @joshmoore convinced me that `sameAs` is probably the best property for these types of external identifiers for now as we already want to support multiple values (ORCID, GitHub) and might want to add more (Twitter, LinkedIn?..). bbdaf9a updates the JSON as well as the HTML template to digest it accordingly

joshmoore (approved):
Big fan!
Screen Shot 2021-04-15 at 16 41 18

I'm still not opposed to the idea of an `@id` or `identifier` but this is a great start. :+1:

PR #473, opened 2021-04-25
Additional event metadata

sbesson:
Fixes a few warnings reported by the [Rich Results Tests](https://search.google.com/test/rich-results?utm_campaign=devsite&utm_medium=jsonld&utm_source=livestream&user_agent=1&url=https%3A%2F%2Fwww.openmicroscopy.org%2Fevents%2Fome-community-meeting-2021%2F)

- Add `eventAttendanceMode/eventStatus` for each `Event`
- Add `organizer.url`
- Add `organizer` for each of the days 2-4
- Add event  registration as `offers`

### DIFF
--- a/_data/ome2021.json
+++ b/_data/ome2021.json
@@ -3,6 +3,7 @@
   "@context": "https://schema.org",
   "@type": "Event",
   "eventAttendanceMode": "https://schema.org/OnlineEventAttendanceMode",
+  "eventStatus": "https://schema.org/EventScheduled",
   "location": {
     "@type": "VirtualLocation",
     "url": "TBD"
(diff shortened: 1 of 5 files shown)


joshmoore (approved):
Wow. Interested to see if this helps the event start showing up in google results. Correction: it wasn't the last time I looked, but now I see:

Screen Shot 2021-04-26 at 07 45 55

The repetition does make me wonder if we're going to need to add some logic around this though.

joshmoore:
p.s. have you seen a way to link the "ome2021" alias to the event?

sbesson:
I saw the repetition. I suspect it's an by-product of the way the metadata is constructed with one overarching `Event` with nested `subEvent` element. I had initially considered creating one `Event` per day and that might help. Don't know if we want to invest into this solution. It's definitely a (good) learning process

> p.s. have you seen a way to link the "ome2021" alias to the event?

You mean https://j.mp/ome2021 ? At minimum I assume that could be represented as `sameAs`

joshmoore:
> You mean https://j.mp/ome2021 ?

I meant more some form of hash tag or alternative name for the event so that it shows up more prominently in search results.

> Don't know if we want to invest into this solution.

Not necessarily. Only if it gets annoying enough :wink:

PR #475, opened 2021-04-29
Add Media Cybernetics to the list of commercial partners

sbesson:


sbesson:
Staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/commercial-partners/. Merging and deploying.

PR #478, opened 2021-05-05
Add speakers to the OME 2021 overview page

sbesson:
As discussed with the organizing committee, this merges the speakers page into the overview page (and adds a redirect for the previous page).

The JSON of the event is updated to use `firstName` and `familyName` and allow the event performers to be sorted by last name.

Also add a banner at the top-level to link to the open registration form


joshmoore (approved):
LGTM

PR #485, opened 2021-05-25
Change break-out sessions into Concluding remarks for Day 4

sbesson:
As discussed earlier today, the last 20min of each session on Day 4 should be used for closing the meeting.


joshmoore (approved):
LGTM

PR #487, opened 2021-05-25
Miscellaneous project updates

sbesson:
- add sentence about logos and trademarks to the footer
- update funding in about page
- update commercial partners
- add anchors to sections of about page and add top-level navigation menu

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -3,7 +3,7 @@
                 <div class="small-12 medium-4 columns">
                     <div class="medium-4"><a href="{{ site.baseurl }}/index.html"><img class="footer-logo" src="{{ site.baseurl }}/img/logos/ome-main-nav.svg" alt="OME logo" /></a></div>
                     <p>&copy; 2005-{{ site.time | date: "%Y" }} University of Dundee &amp; Open Microscopy Environment. <a href="http://creativecommons.org/licenses/by/4.0/" target="_blank">Creative Commons Attribution 4.0 International License</a></p>
-                    <p>OME source code is available under the <a href="http://www.gnu.org/copyleft/gpl.html">GNU General public license</a> or more permissive open source licenses, or through commercial license from <a href="https://www.glencoesoftware.com" target="_blank">Glencoe Software</a>.</p>
+                    <p>OME source code is available under the <a href="http://www.gnu.org/copyleft/gpl.html">GNU General public license</a> or more permissive open source licenses, or through commercial license from <a href="https://www.glencoesoftware.com" target="_blank">Glencoe Software Incl.</a><br/>OME, Bio-Formats, OMERO and their associated logos are trademarks of <a href="https://www.glencoesoftware.com" target="_blank">Glencoe Software Incl.</a>, which holds these marks to protect them on behalf of the OME community.</p>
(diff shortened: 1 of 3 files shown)


sbesson:
See https://snoopycrimecop.github.io/www.openmicroscopy.org/about/ and https://snoopycrimecop.github.io/www.openmicroscopy.org/commercial-partners/

_includes/footer.html
```diff
@@ -3,7 +3,7 @@
...
                     <p>&copy; 2005-{{ site.time | date: "%Y" }} University of Dundee &amp; Open Microscopy Environment. <a href="http://creativecommons.org/licenses/by/4.0/" target="_blank">Creative Commons Attribution 4.0 International License</a></p>
-                    <p>OME source code is available under the <a href="http://www.gnu.org/copyleft/gpl.html">GNU General public license</a> or more permissive open source licenses, or through commercial license from <a href="https://www.glencoesoftware.com" target="_blank">Glencoe Software</a>.</p>
+                    <p>OME source code is available under the <a href="http://www.gnu.org/copyleft/gpl.html">GNU General public license</a> or more permissive open source licenses, or through commercial license from <a href="https://www.glencoesoftware.com" target="_blank">Glencoe Software Incl.</a><br/>OME, Bio-Formats, OMERO and their associated logos are trademarks of <a href="https://www.glencoesoftware.com" target="_blank">Glencoe Software Incl.</a>, which holds these marks to protect them on behalf of the OME community.</p>
```
joshmoore, line 6: maybe one extra `<br/>`?

about/index.html
```diff
@@ -3,24 +3,21 @@
 title: About Us
 meta_description: OME produces open tools to support data management for microscopy. Learn more here.
 funding:
-    - grant_name: "BBSRC BBR Award"
```
joshmoore, line 6: Is this "previous funding"?

about/index.html
```diff
@@ -101,6 +98,17 @@
     - group2_name: "Electron Microscopy Public Image Archive (EMPIAR)"
       group2_link: "https://www.ebi.ac.uk/pdbe/emdb/empiar/"
 ---
+        <div class="menu-subnav" id="about-menu">
```
joshmoore, line 101: Nice!

joshmoore (commented):
Few minor comments/questions

_includes/footer.html
sbesson: I.e. further separate the two paragraphs? I found this was at the detriment of the balance of the footer since the left column becomes much larger than the other ones.

Thoughts about moving the logo and/or the sitemap and/or the version to other columns?

about/index.html
sbesson: Definitely previous funding. Would you keep a no-op dictionary for historical funding?

As an aside, I looked into https://schema.org/Grant which would be an interesting way forward and would likely capture the main concepts. Two caveats: this element is not integrated in the official schema.org spec yet and I have not found a good way to capture active/previous concepts.

_includes/footer.html
joshmoore: If you tried it out, don't worry about it.

about/index.html
joshmoore: > Would you keep a no-op dictionary for historical funding?

Yes. Seems a shame to lose it.

sbesson: 95d469f

joshmoore (approved):
:+1:

PR #492, opened 2021-06-01
list demo

jburel:
Add entries for demo

More to be added

### DIFF
--- a/_includes/ome2021_program_agenda.html
+++ b/_includes/ome2021_program_agenda.html
@@ -67,7 +67,7 @@
     <ul text-align="left">
     {% for evt in subevent.subEvent %}
       <li>
-      <a name="{{ subevent.name }}"></a><h5>{{ evt.name }}</h5>
+      <h5>{{ evt.name }}</h5>
       <p>
       {% if evt.description %}{{ evt.description }}<br/>{% endif %}
       {% for performer in evt.performer %}{{ performer.givenName }} {{ performer.additionalName }} {{ performer.familyName }}, <span class="ome-green italic">{{ performer.affiliation.name }}</span><br/>{% endfor %}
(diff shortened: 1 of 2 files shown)


_includes/ome2021_program_agenda.html
```diff
@@ -64,7 +64,21 @@ <h5>{{ session.name }}</h5>
     {% if subevent.description %}{{ subevent.description }}<br/>{% endif %}
     {% for performer in subevent.performer %}{{ performer.givenName }} {{ performer.additionalName }} {{ performer.familyName }}, <span class="ome-green italic">{{ performer.affiliation.name }}</span><br/>{% endfor %}
     </p>
-  </div>
+    <ul text-align="left">
+    {% for evt in subevent.subEvent %}
+      <li>
+      <a name="{{ subevent.name }}"></a><h5>{{ evt.name }}</h5>
```
sbesson, line 70: Interestingly on Desktop this creates an unwanted extra line
Screenshot 2021-06-02 at 08 57 32
but it looks much better on mobile.

Is the anchor necessary at all as they will all likely have the same name since they are linked to the parent `subevent` name?
If we need it, could you try moving the `<a>` element inside the `<h5>`?

sbesson (changes requested):
One rendering issue. Otherwise, looks good from my side

_includes/ome2021_program_agenda.html
jburel: Which browser? I don't see the problem.I am using Chrome

sbesson: Interesting, I cannot reproduce on Chrome either. The issue shows up on Firefox

sbesson (approved):
Now looks good. Merging as more changes are coming on Day 1 and the re-usage of the sub-event for concurrent break-out sessions.

Running the JSON through https://search.google.com/test/rich-results, the only required attribute missing from these new demo events is a `startDate`. I assume we should just copy the `startDate` (and `endDate`) from the parent event but this can happen in a separate PR.

PR #495, opened 2021-06-02
Day1

jburel:
Adjust time and schedule
This also removed the targeted audience of Session 2

### DIFF
--- a/_includes/ome2021_program_agenda.html
+++ b/_includes/ome2021_program_agenda.html
@@ -10,8 +10,7 @@
     <h5>Please Note</h5>
     <p>
         This schedule is subject to change and will be updated periodically.<br/>
-        The schedule will include pre-recorded videos organised by topic followed by group discussions.<br/>
-        All times are in UK - British Summer Time (GMT+1).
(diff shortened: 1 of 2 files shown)


_includes/ome2021_program_agenda.html
```diff
@@ -10,7 +10,8 @@ <h2>{{ event.name }}</h2>
     <h5>Please Note</h5>
     <p>
         This schedule is subject to change and will be updated periodically.<br/>
-        The schedule will include pre-recorded videos organised by topic followed by group discussions.
+        The schedule will include pre-recorded videos organised by topic followed by group discussions.<br/>
+        All times are in UK - British Summer Time (GMT+1).
```
sbesson, line 14: I think I had moved this sentence under the `Schedule` section (as this is the only one which indicates day times)

jburel: I missed it

sbesson (approved):
Looks good. The `endDate` for the demos can probably be updated but that does not impact the program and I can handle this as a follow-up.

PR #496, opened 2021-06-03
Playlist updates

sbesson:
As discussed at this morning's call:

- adds the playlist with the Community Meeting introductions to Day 1
- adds the video count and the playlist length to the name field of the VideoObject (thanks to https://ytplaylist-len.herokuapp.com/)
- updates the `endDate` of the demos
- minor fixes (organizers' affiliations, preliminary -> final)

### DIFF
--- a/_data/ome2021.json
+++ b/_data/ome2021.json
@@ -96,7 +96,7 @@
       "workFeatured": [
         {
           "@type": "VideoObject",
-          "name": "Community Meeting Introductions",
+          "name": "Community Meeting Introductions (3 videos, playlist length: 16min)",
           "uploadDate": "2021-06-03",
           "description": "Introduction to the OME 2021 Community Meeting",
           "contentUrl": "https://www.youtube.com/playlist?list=PL-oOCWFUMH50iXx_0ERRt0URhnx14fMJ0",
@@ -105,7 +105,7 @@
         },
         {
           "@type": "VideoObject",
-          "name": "Demo Introductions",
+          "name": "Demo Introductions (2 videos, playlist length: 7min)",
           "uploadDate": "2021-05-28",
           "description": "Introduction to the demos of Day 1",
           "contentUrl": "https://www.youtube.com/playlist?list=PL-oOCWFUMH51ACy8QhTYc4ppaoICJQDU_",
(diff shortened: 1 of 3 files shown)


joshmoore:
Screen Shot 2021-06-03 at 15 39 48

Nice. and kudos to @pwalczysko for the suggestion.

_data/ome2021.json
```diff
@@ -96,7 +96,16 @@
       "workFeatured": [
         {
           "@type": "VideoObject",
-          "name": "Demo Introductions",
+          "name": "Community Meeting Introductions (3 videos, playlist length: 16min)",
```
joshmoore, line 99: Aha, but hardcoded. Perhaps you can paste how you used the app so we can repeat what you did?

index.html
```diff
@@ -84,7 +84,7 @@ <h1 class="show-for-small-only hero-main-header"><img src="{{ site.baseurl }}/im
                     <h1 class="small-12">OME 2021 Community Meeting</h1>
                     <h5 class="small-10 small-offset-1 medium-4 medium-offset-4 bold text-shadow-light">Join us from June 08th to June 11th 2021 for our virtual community meeting.</h5>
                     <br>
-                    <a href="{{ site.baseurl }}/events/ome-community-meeting-2021/" class="large button sites-button btn-red">View the preliminary programme</a>
+                    <a href="{{ site.baseurl }}/events/ome-community-meeting-2021/" class="large button sites-button btn-red">View the final programme</a>
```
joshmoore, line 87: Ok for this to go live now?

joshmoore:
Generally looks fine to me. Few questions/points, no blockers.

_data/ome2021.json
sbesson: It looks like there is a way using the Youtube API but I did not really have the time to investigate.

The workflow is pretty straightforward albeit manual: go to the URL, paste the full URL or the ID of the playlist and click on `Get Length`
Screenshot 2021-06-03 at 14 43 07

index.html
sbesson: Worth confirming with the organizing committee but I would say the program (or at least the times) are as final as can be. The only difference might be more demos/break-out sessions.

PR #499, opened 2021-06-04
Split demo in slots

jburel:
Update day1 demos section


jburel:
Merging and deploying before sending emails to participants doing demo

PR #500, opened 2021-06-04
move schedule section

jburel:
cc @joshmoore


jburel:
Display matches the layout used last year
Screenshot 2021-06-04 at 16 35 37

PR #502, opened 2021-06-04
remove entry

jburel:
Update demo list


jburel:
Screenshot 2021-06-04 at 21 59 02

PR #505, opened 2021-06-09
Add keynote Youtube video to JSON object and embed within Day 1

sbesson:
The video is still up for discussion but this should provide a first template for embedding videos recorded during the meeting. Could also be applied to individual demo recordings and other days


sbesson:
Deployed at https://snoopycrimecop.github.io/www.openmicroscopy.org/events/ome-community-meeting-2021/day1/

sbesson:
Looks good. Let's deploy and iterate with new content

PR #508, opened 2021-06-09
Add details for break-out sessions on Day 4

sbesson:
Should we include a break-out session for a round-table following the Q&A session? /cc @jburel


jburel:
We have been using the main room as the buffer room for other session. so it might make sense to have a round table
the problem is now that we have already 3 rooms in the morning already

PR #510, opened 2021-06-14
Fix link to explore

pwalczysko:
Found out during a demo on OME2021. The link "Give OMERO a quick try" on https://www.openmicroscopy.org/omero/institution/ is broken, which is not detectable either on local build nor on snoopy (only on the production site).

### DIFF
--- a/omero/institution/getting-started.html
+++ b/omero/institution/getting-started.html
@@ -34,7 +34,7 @@ meta_description: To help you get started installing OMERO for your institution,
                 </ul>
                 <h5>Read and watch about OMERO or try it to get a better idea.</h5>
                 <ul class="dev-ops-links">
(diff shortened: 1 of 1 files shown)


pwalczysko:
built at https://snoopycrimecop.github.io/www.openmicroscopy.org/omero/institution/getting-started.html

joshmoore (approved):
LGTM

joshmoore:
Anything else to roll into a release?

sbesson (approved):

sbesson:
The only other thing I can think of would be the removal of the banner since the community meeting is now done.
Is there anything else we want to promote? And/or do we want to add an event for the upcoming ELMI workshops at https://www.openmicroscopy.org/events/ ? cc @jburel @francesw

francesw:
Yes, can include the ELMI workshops.  There's also the microscopy workshop at EBI in July, if we want to include that too.

PR #511, opened 2021-06-15
fix youtube link ome2021

will-moore:
Replace incomplete Claire Stoffel video with full length one.


sbesson (approved):
That's great. Since this is the video of the re-run of the demo on Friday, the only question is whether it should be moved under the corresponding event or whether we would rather have all the demo videos on the Day 1 page. Otherwise LGTM

will-moore:
Don't know. Could put it there and *add* a link from here on day1 (and leave the existing video here)?

sbesson:
I have no strong preference. I'll get this merged as it stands. If someone feels strongly otherwise, we can revisit

PR #513, opened 2021-06-17
Fix the missing bar between the last and one but last event

pwalczysko:
@sbesson 

One minor thing which I found out just before tagging. 

Can you pelase have a look ?


sbesson (approved):

pwalczysko:
@sbesson thank you

PR #515, opened 2021-06-21
Migrate OME blog content into main OME website

sbesson:
For historical reasons, the OME blog is currently externalized and deployed at https://blog.openmicroscopy.org/ and linked from the OME website via the  `New > Blog` menu. 

Following the migration of the OME website from Plone to Jekyll a few years ago, the idea of migrating the blog items back into the website has been mentioned several times but never turned into action. A benefit of this unification is the re-usage of the styling/assets as well as unifying our resources to a single repository for adding a blog post/announcement/....

This PR attempts to merge the blog content into the website. Once deployed in production, we can update https://github.com/ome/blog and Jekyll redirect plugin to redirect the existing blog URLs to the new location. 

The blog commits were created using

```
rm -rf /tmp/blog/ && git clone https://github.com/ome/blog /tmp/blog && cd /tmp/blog

WHITELIST='^(_posts/.*|_data/.*|images/.*|_includes/caption.html)$'
git filter-branch -f --prune-empty --index-filter "git ls-files | egrep -v \"$WHITELIST\" | tr "\\\\n" "\\\\0" | xargs -0 -r git rm -f --cached --"

tmpfile=`mktemp -t splitgitXXXXXX`
chmod 755 $tmpfile
echo '#!/usr/bin/env ruby
old_parents = gets.chomp.gsub("-p ", " ")
new_parents = old_parents.empty? ? [] : `git show-branch --independent #{old_parents}`.split
puts new_parents.map{|p| "-p " + p}.join(" ")' > $tmpfile
for i in $(seq 3); do
  git filter-branch -f --prune-empty --parent-filter $tmpfile HEAD
done

mkdir blog && git mv _posts blog && git add blog && git commit -m "Move all blog posts under blog subfolder"
```

and merged using `--allow-unrelated-histories` option.

A few initial design questions need to be settled before making progress:

- layout: d3e44f5 and f6fe334 use a layout similar to https://www.glencoesoftware.com/blog/ where the blog posts are prefixed with `blog` and managed independently from the Announcements. An alternative would be to add blog posts under `_posts` so that they would appear under `Announcements` and possibly use tag/categories to differentiate?
- tag/categories: the existing usage of blog categories are encoded into the URL creating URLs of type e.g. `www.openmicroscopy.org/blog/data-model/future-plans/2016/06/20/shape-transforms.html`. This departs from the announcements which are under `www.openmicroscopy.org/YYYY/MM/DD/title.html`
- images were stored under `images` under the blog vs `img` under the OME website. This probably requires unification e.g. `img/blog`?


joshmoore:
(Note the first line of your description should describe blog.o.org not www.o.org)


Generally :+1: but:
* News > Blog still redirects to blog.o.org
* I'm in support of having categories, but they don't need to be in the URL (I was never a big fan of that)
* I'm a tiny bit worried about dropping RSS. Can we redirect https://blog.openmicroscopy.org/atom.xml somewhere useful?

sbesson:
So keep the `/blog` namespace but use have e.g. `https://www.openmicroscopy.org/blog/2016/06/20/shape-transforms.html` ? I will look into categories -> tags as I suspect this is the most straightforward way.

I will also try to resurrect the atom feed (might be something we want for announcements & events too?)

sbesson:
> I'm in support of having categories, but they don't need to be in the URL (I was never a big fan of that)

It just occured to me that it should be possible to have the blog posts stored under the root `_posts` with `category: blog` and `tags: file-formats community`. This should suffice to have the URL prefixed with `/blog/` only and  have a blog landing page with all posts. It might also cause the blog posts to be listed under https://www.openmicroscopy.org/announcements/ which might or might not be what we want.

joshmoore:
No immediate objection to having blog items (or even others) under "announcements" _unless_ that becomes the _only_ RSS feed which might be suboptimal.

joshmoore:
* My RSS client can parse the atom.xml
 * The link in the menu works.
 * 
Are the big remaining issues categories (I didn't find them) and maybe some styling?

sbesson:
Definitely planning to work on the styling next if we agree with the layout proposal.

I am still thinking on the categories/tags concept. Using this issue to capture the current state of this PR but also brainstorming on potential usage:

- a post can have either no category or have `category: blog`. In the first case, it will only be listed under `News > Announcements`. In the latter case, it will also be listed under `News > Blog`
- There are also two other de facto posts categories managed outside `_posts`: `minutes` and `events`.
- as per the above, I would say a category should a dedicated entry under the News menu with an index page and optionally a dedicated Atom/RSS feed.
- in addition to categories,  posts can have tags. The current proposed tags are inherited from OME Blog and include `file-formats`, `data-model`, `future-plans`, `community`, `tools`, `tech-issues`
- as we will put redirect from the previous URLs, there is no requirement to keep all these tags
- there is no exposition of the tag in the UI. Assuming this list does not grow infinitely, it looks doable to have a list of tags e.g. at the top of `Announcements` linking to landing index pages for all posts of a given tag e.g. under `https://www.openmicroscopy.org/announcements/file-formats` or `https://www.openmicroscopy.org/announcements/tag/file-formats`
- the list of tags could be expanded to also support the announcements posts e.g. `omero-insight`, `bio-formats`, `omero-server` could easily be added as tags.

snoopycrimecop:
Conflicting PR. Removed from build [WEBSITE-push#304](https://latest-ci.openmicroscopy.org/jenkins/job/WEBSITE-push/304/). See the [console output](https://latest-ci.openmicroscopy.org/jenkins/job/WEBSITE-push/304/consoleText) for more details.
Possible conflicts:
  - Upstream changes
    - build.sh

--conflicts

sbesson:
While https://github.com/ome/www.openmicroscopy.org/pull/515#issuecomment-866712652 discusses a more ambitious plan, I would try to define a more focused scope to get this PR merged, archive https://github.com/ome/blog and use this website to publish new blog posts (as several are already planned). The tag filtering can be captured as a separate usability issue.

The last few commits:

- unify the permalink of all posts e.g. https://snoopycrimecop.github.io/www.openmicroscopy.org/2020/12/01/zarr-hcs.html or https://snoopycrimecop.github.io/www.openmicroscopy.org/2021/08/23/bio-formats-6-7-0.html
- add some background images for the blog category index and the minutes category
- update the post layout to reflect the category 

Deployed at http://snoopycrimecop.github.io/www.openmicroscopy.org/blog/ for review.

joshmoore:
I don't understand the potential (future) tags/categories fixes to know if we run the risk of needing to break URLs, but worst case, we maintain static registry pages for each category:

```
/blog/categories/file-formats.html <-- we list the other URLs.
```

Big :+1: for getting this through. (2021 is sorely lacking in blog posts :wink: )

sbesson:
I think the critical assumption for the URL stability is the one expressed in https://github.com/ome/www.openmicroscopy.org/pull/515#issuecomment-865758596

> I'm in support of having categories, but they don't need to be in the URL (I was never a big fan of that)

Unless we decide to revert this decision, then there should be no breaking change to the individual post permalinks.

I'll get this listed for a final round of comments tomorrow. Once merged and deployed on the OME website, I'll work on redirecting the legacy blog URLs to their new location.

sbesson:
Briefly mentioned this morning, no objection has been raised against this. Merging and updating the website.

PR #526, opened 2021-08-03
Temporarily use custom User-Agent for htmlproofer

sbesson:
The GitHub actions have been broken for the last 10 days due to an impossibility to validate a Science DOI. This seems to be related to the upstream website blocking user agents - see https://github.com/gjtorikian/html-proofer/issues/648 for the related conversation.

This PR passes a custom Mozilla agent to workaround the problem. The Science website has been separately contacted (https://www.sciencemag.org/about/contact-us) to ask about adding the htmlproofer to the allowed list. Hopefully this means this workaround can be reverted in the near future.


joshmoore (approved):
Makes sense based on recent interactions with image.sc, even if `--typhoeus-config='{"headers" :` is quite the surprise ;)

PR #535, opened 2021-09-14
bump insight version to 5.6.2

jburel:
Bump version


sbesson (approved):
Is it worth an announcement? Or just release as it is to fix the downloads page?

jburel:
I will push a post with the new version since we started to do it again from 5.6.0

PR #541, opened 2021-10-04
Idr logo

jburel:
Add row for IDR logo
Currently it is a placeholder. To be modified when the logos/zip are ready

cc @sbesson @joshmoore


sbesson:
Added the IDR logo & logomark. Now deployed at https://snoopycrimecop.github.io/www.openmicroscopy.org/artwork/. A few comments:

- limited to color-only in SVG/PNG format for now
- the SVG version of the logomark has been derived from the logotype + logomark using Inkscape
- created a PNG version of the  the logotype + logomark (`logo`) and logomark using `mogrify -format png`

jburel:
I think it is find as it is. We can improve it later on.

PR #545, opened 2021-11-02
Regenerate IDR PNG logos with transparent background

sbesson:
Also generate a white version of the logo as used in presentations

The SVG files are unchanged but the PNG files have been regenerated with transparent background (to facilitate the embedding e.g. in presentations) and a white version of the PNG, similar to the ones used in the header of the standard IDR slides has been created.

NB: I am parallely starting to look into managing the sources and the generation of the raster image (see https://github.com/sbesson/artwork and https://github.com/sbesson/artwork/actions/runs/1414246457). I currently envision modifications to the logos would be release assets of the artwork repository and could be consumed in this repository (and others). I am happy to take any suggestions on tying (or not) both efforts.


sbesson:
https://github.com/sbesson/artwork is probably now in a state where it can be reviewed:

- sources (SVG versions) are kept under version control
- PNG versions are generated with a transparent background (ImageMagick 7/OSX)
- a build script generates all PNGs and creates 3 zips (OME, OME products, IDR)
- GHA runs the builds and creates artifacts for each PR and create a release with assets on tag - see https://github.com/sbesson/artwork/releases/tag/2021.11.16

If we are happy with this approach, I'd propose to migrate the repo to the `ome` org, recreate a release to test the process and update the zips in this repository with official release assets.

joshmoore:
* Happy with the approach
* The PNGs look good
* Outwith this PR:
  * I could imagine some of them with white or black backgrounds 
  * I could almost imagine as well having them unzipped for hot-linking but an alternative would be to have a single page that consumes all of them in some re-usable display

sbesson:
Cheers. I'll leave another day for comments and then go ahead with the repo transfer.

>Outwith this PR:
>    I could imagine some of them with white or black backgrounds
>    I could almost imagine as well having them unzipped for hot-linking but an alternative would be to have a single page that consumes all of them in some re-usable display

Proposing to capture these as issues/RFE. For the second one, I assume https://www.openmicroscopy.org/artwork/ could become this page and give access either to bundles (like https://github.com/logos) and/or individual files (like https://git-scm.com/downloads/logos). There might be a case for generating another release artifact with all the logos (SVG + PNG) that could easily be extracted here.

joshmoore:
> Proposing to capture these as issues/RFE.

:+1:

Also: https://calibreapp.com/blog/compress-images-in-prs

joshmoore:
> I'll leave another day for comments and then go ahead with the repo transfer.

:+1:

jburel:
Sorry I completely missed this PR. Things look good. Include in the next deployment?

sbesson:
Re-reading the thread, I think there are two issues which conflated and put this  on hold:

- the generation of PNG with transparent background
- the management of the SVG sources for all our logos/icons

The latter is still  waiting on a decision regarding the licensing.

The former should be ready to go although looking at it again, it raises the question of whether we would want to generate both types of PNG (white background and transparent background) and use a suffix to differentiate them.

joshmoore:
Sorry for letting this slip. Concentrating on the former, my ordering would be:

 * All (e.g. [here](https://github.com/zarr-developers/zarr-logo/releases/tag/2022-04-28))
 * Just transparent
 * Whatever else

But since this is currently in place with the second best choice, I'm happy to get it in and we can go from there. Thanks, @sbesson!

sbesson:
Despite being still involved in the IDR project, I don't feel I am the best placed to make proposals on the IDR branding. So I am going to close this but if this is of value, feel free to re-use the updated binaries.

PR #547, opened 2021-11-17
remove banner

jburel:
Notice while checking the status of the website


sbesson (approved):
No objections from my side to removing this banner as it's been a month and we have released another OMERO.web release (is it worth an announcement on the OME website @will-moore ?)

will-moore:
I don't think the last release really needs an announcement, unless it's our policy to always add one?

jburel:
we do not add one every time expect for application like insight, BF.

PR #550, opened 2021-12-02
Add citation for 2021 OME-NGFF paper

sbesson:
Now that the paper is out, add a citation for https://doi.org/10.1038/s41592-021-01326-w using a new heading

NB: I considered going the road (rabbit hole?) of converting the content of this page into data files using https://schema.org/ScholarlyArticle. Unsure whether anyone has strong views on the value of doing that.

### DIFF
--- a/citing-ome/index.html
+++ b/citing-ome/index.html
@@ -43,6 +43,21 @@ meta_description: Learn how to properly cite OME work in your own publications.
         <hr class="whitespace">
         <div class="row">
             <div class="small-12 medium-4 columns">
+                <h3>OME-NGFF</h3>
+            </div>
+            <div class="small-12 medium-8 columns">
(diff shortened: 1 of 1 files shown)


pwalczysko:
Looks good on https://snoopycrimecop.github.io/www.openmicroscopy.org/citing-ome/

I suppose the PMCID link is missing because the paper is not listed by PMC ? (at least I did not find it using the search for the title)

sbesson:
Yes the PMC ID is not available yet. If the submission process comes form the journal, there is possibly a period of embargo which depends on the submitter (up to 12 months) - https://www.ncbi.nlm.nih.gov/pmc/about/submission-methods/

pwalczysko (approved):

PR #551, opened 2021-12-08
Bio-Formats 6.8.0 updates

dgault:
Updates to bump the Bio-Formats version and add the release announcement


sbesson:
https://github.com/ome/www.openmicroscopy.org/blob/master/_data/bio-formats.json should also be reviewed to include the recent contributors

dgault:
Builds failures due to 6.8.0 docs not yet deployed

sbesson (changes requested):
Can the latest release on the top-level index.html page also be updated?

Otherwise https://snoopycrimecop.github.io/www.openmicroscopy.org/2021/12/09/bio-formats-6-8-0.html looks great!

PR #555, opened 2021-12-13
Update for Log4Shell (CVE-2021-44228)

joshmoore:
* Add a landing blog post following style of the 2015 "java-issue"
* Add a 2021-SV4 for loci_tools.jar
* List both on the advisories index
* Activate the banner

### DIFF
--- a/_posts/2021-12-13-log4j-issue.md
+++ b/_posts/2021-12-13-log4j-issue.md
@@ -1,7 +1,7 @@
 ---
 layout: post
 title: CVE-2021-44228 ("Log4Shell") Assessment
-categories: tech-issues
+categories: blog
 ---
 
 Major news carriers have been reporting recently on log4j vulnerabilities in
--- a/_posts/2021-12-13-log4j-issue.md
+++ b/_posts/2021-12-13-log4j-issue.md
@@ -1,0 +1,15 @@
+---
+layout: post
+title: CVE-2021-44228 ("Log4Shell") Assessment
+categories: tech-issues
+---
+
+Major news carriers have been reporting recently on log4j vulnerabilities in
+Java applications. Both the OME team in Dundee as well as Glencoe Software have
+evaluated the libraries used by OMERO.server and OMERO.insight. We can say with
+confidence that OMERO and OMERO Plus are not vulnerable, as they do not utilize log4j
+and no immediate action that needs to be taken.
+
+The only remaining use of log4j in OME software is the `loci_tools.jar` and we
+would suggest moving to the `bioformats_package.jar`. This has been published
+as [2021-SV4 log4j in loci_tools.jar]({{ site.baseurl }}/security/advisories/2021-SV4).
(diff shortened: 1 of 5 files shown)


joshmoore:
Doh, I missed during the migration, @sbesson, the fact that we needed to add `.html` (https://github.com/ome/www.openmicroscopy.org/pull/555/commits/9fa269fb0d8e0fef1abb8de85b935b7be01fd0a9)

joshmoore:
Failures are now related to:
```
http://www.openmicroscopy.org/Schemas/.*
```

sbesson:
> Doh, I missed during the migration, @sbesson, the fact that we needed to add .html (9fa269f)

Yes, as part of the migration, we decided to treat the blog content like posts, with blog becoming a category. Unfortunately, we had a split horizon in terms of permalinks between the old OME blog and the website which means this changed the blog permalinks to use the `.html` suffix - https://github.com/ome/www.openmicroscopy.org/blob/master/_config.yml#L29

I also prefer the `.html`-less version and we did something similar for advisories on the past - https://github.com/ome/www.openmicroscopy.org/pull/115. This would require updating the permalink either for all posts or just a category, adding `redirect_from` for backwards-compatibilty. Should we capture as a RFE?

_posts/2021-12-13-log4j-issue.md
```diff
@@ -0,0 +1,18 @@
...
+
+Major news carriers have been reporting recently on log4j vulnerabilities in
+Java applications. Both the OME team in Dundee as well as Glencoe Software have
+evaluated the libraries used by OMERO.server and OMERO.insight. We can say with
+confidence that OMERO and OMERO Plus are not vulnerable, as they do not utilize log4j
+and no immediate action that needs to be taken.
+
+The only remaining use of log4j in OME software is in the
+[deprecated](https://imagej.net/pipermail/imagej-devel/2015-March/002513.html)
+`loci_tools.jar` which we will now be removing. The `bioformats_package.jar` uber-jar
```
sbesson, line 15: To be discussed but can we be authoritative and say "which we will now be removing in the next release of Bio-Formats"? /cc @dgault @melissalinkert

sbesson (changes requested):
The submodule change should be reverted.

One question about the `loci_tools` timeline for the @ome/formats team. Otherwise, this assessment also matches my understanding of our (limited) usage of log4j

From a quick audit of https://github.com/search?q=org%3Aome+log4j&type=Code, after filtering the archived repositories (like pydoop-features):
- the `loci_tools` uber JAR is certainly the biggest consumer
- in Bio-Formats, the other usage of `log4j` is in the Bio-Formats MATLAB logging tests. I assume either this is covered by the previous recommendation (i.e. switch to the logback uber-JAR) or is a MATLAB problem wrt which JARs are shipped in their distirbutions
- a few places in downstream components like [omero-py](https://github.com/ome/omero-py/blob/94b4be90f85f8ccaef02ceebc75320cf1b70db38/src/omero/java.py#L59)  and [minimal-omero-client](https://github.com/ome/minimal-omero-client/blob/67800a9b5e9a6dd73562aab361ea8be45f676d96/build.gradle#L9) refer to log4j configuration files. It might be worth reviewing these places & clarifying their intent in this assessment.

joshmoore:
Pushed a fix, @sbesson. We can make a decision after the formats meeting, but if we can't be more definitive today, I'd suggest we get this released ASAP. (Note: bfmatlab.zip ships logback)

sbesson (approved):

PR #558, opened 2021-12-17
Update favicon from https://realfavicongenerator.net/

joshmoore:
see:
 * https://forum.image.sc/t/ome-favicon-discrepancy/61080
 * https://github.com/imagej/imagej.github.io/issues/214


joshmoore:
The checker is not completely happy with snoopycrimecop.github.io/www.openmicroscopy.org/ because several of the files are expected in the root. That will be fixed on deployment.

Note the use of the `?v=2` suffix in order to trigger re-caching of the ico.

sbesson (approved):
Let's get this deployed and we can use the same strategy everywhere.

joshmoore:
:+1:

sbesson:
https://www.openmicroscopy.org/ now passes the checks on https://realfavicongenerator.net/favicon_checker?protocol=http&site=www.openmicroscopy.org%2F#.Ybx62yynxqs

PR #560, opened 2022-01-05
Review usage of NGFF vs OME-NGFF

sbesson:
Raised post-standup today, this aligns with the semantics used in https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8648559/ where NGFF is used to characterize the cloud-optimized formats like Zarr and typically balanced versus TIFF or HDF5 while OME-NGFF is used to talk about the format and the set of specifications built by OME and partners allowing to use NGFFs for storing, acessing and exchanging bioimaging data.

This implicitly proposes to rename the blog post URL to https://www.openmicroscopy.org/2021/12/16/ome-ngff.html (with a redirection from the former URL). For build reasons, the new URL will need to exist live to pass the linkchecker so 
541e386 first introduces a redirect from `ome-ngff.html` to `ngff.html`. Once deployed, a second change can swap the pages so that `ngff.html` redirects to `ome-ngff.html `

Staged at https://snoopycrimecop.github.io/www.openmicroscopy.org/2021/12/16/ngff.html for review

### DIFF
--- a/_posts/2021-12-16-ngff.md
+++ b/_posts/2021-12-16-ngff.md
@@ -4,7 +4,7 @@ title: OME-NGFF 2021 updates
 tags: file-formats community
 categories: blog
 redirect_from:
-  - /2021/12/16/ngff.html
+  - /2021/12/16/ome-ngff.html
 
 ---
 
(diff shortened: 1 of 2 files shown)


joshmoore:
:+1: As I mentioned this morning, this seems to match my usage of `OME-NGFF` vs `NGFF(s)`

PR #561, opened 2022-01-06
Rename blog post URL as ome-ngff

sbesson:
See comment in https://github.com/ome/www.openmicroscopy.org/pull/560


sbesson:
See https://snoopycrimecop.github.io/www.openmicroscopy.org/2021/12/16/ome-ngff.html with https://snoopycrimecop.github.io/www.openmicroscopy.org/2021/12/16/ngff.html now redirecting to this page

sbesson:
No objection raised in standup

PR #563, opened 2022-01-13
Bio-Formats 6.8.1 updates

dgault:
Updates to bump the Bio-Formats version and add the release announcement


sbesson (commented):
All looks good. Should the announcement be posted today rather than tomorrow?

sbesson (approved):

PR #566, opened 2022-01-26
Add security advistories to the footer

joshmoore:
Up the visibility in addition to News > Security.

(Also fix a blogs link

Screen Shot 2022-01-26 at 10 54 42

Screen Shot 2022-01-26 at 10 58 17
)

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -34,6 +34,7 @@
                         <li><a href="{{ site.baseurl }}/docs/">Documentation</a></li>
                         <li><a href="https://omero-guides.readthedocs.io/">User Guides</a></li>
                         <li><a href="https://www.openmicroscopy.org/forums">Forum</a></li>
+                        <li><a href="https://www.openmicroscopy.org/security">Security Advisories</a></li>
                         <li class="hide-for-large-only">&nbsp;</li>
                     </ul>
                 </div>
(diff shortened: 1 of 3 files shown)


sbesson (approved):

_includes/footer.html
```diff
@@ -34,6 +34,7 @@ <h6>Knowledge Base</h6>
                         <li><a href="{{ site.baseurl }}/docs/">Documentation</a></li>
                         <li><a href="https://omero-guides.readthedocs.io/">User Guides</a></li>
                         <li><a href="https://www.openmicroscopy.org/forums">Forum</a></li>
+                        <li><a href="https://www.openmicroscopy.org/security">Security Advisories</a></li>
```
sbesson, line 37: Should this link to https://www.openmicroscopy.org/security/advisories/ directly then?

joshmoore: Yeah, I wasn't sure either. There's not a link from the list (which is certainly more of a "knowledge base") back to the top, so this seemed safer.

sbesson: Agreed. Interesting individual advisories have a back link to the list of advisories - https://www.openmicroscopy.org/security/advisories/2021-SV4/. 

So it feels like having a `back to Security` in https://www.openmicroscopy.org/security/advisories/ would be valuable either way. It does not need to happen right in this PR though.

joshmoore:
Ok. I'm going to move forward with this and we can see if we don't need to remove one or the other of them.

PR #567, opened 2022-01-26
Link directly to advisories page and add back reference to Security page

sbesson:
Minor and not-urgent suggestion addressing the concerns raised in https://github.com/ome/www.openmicroscopy.org/pull/566#discussion_r792480575


sbesson:
This got inadvertently merged and deployed live - see https://www.openmicroscopy.org/security/advisories/

PR #602, opened 2022-07-22
bump omero-matlab version

jburel:
Bump version to match https://github.com/ome/omero-matlab/releases/tag/v5.5.6


jburel:
see https://snoopycrimecop.github.io/www.openmicroscopy.org/omero/downloads/

PR #613, opened 2022-10-12
Bio-Formats 6.11.0 updates

dgault:
Updates to bump Bio-Formats version and add release announcement and new contributors.
Will fail and wont be ready for full review until BF artifacts have been released


joshmoore (approved):
Looks like a ton of work :wink:

👍🏽

PR #620, opened 2022-10-25
temporarily remove the external checks

jburel:
remove the external check
I have tried other options like ``--swap-urls`` but without much luck.
When we have the site deployed we can remove the flag


joshmoore (approved):
Unfortunate, but I'm not sure of any other solutions. Let's go with this and then reverse after the release.

:+1:

PR #621, opened 2022-10-25
Check external links

jburel:
* Reverse flag introduced in 7ae8654cd8d63428ae94dc3054f01df60ea52bb5
* Fix broken external link

Check that the build is green


will-moore (approved):
change looks good and build is green 👍

PR #627, opened 2022-12-13
Fix name

jburel:
Build number has been dropped from artifacts on github


sbesson (approved):
Matches the artifact layout in OMERO 5.6.6

PR #632, opened 2023-01-18
Insight 5.8.0

jburel:
Add entry for the latest insight version


pwalczysko (approved):
Makes sense, builds okay, also checked the downloads page, it shows 5.8.0

PR #638, opened 2023-02-17
Fix docker command similar to idr.op...py.org

pwalczysko:
Fixes the docker command in line with the working version on https://github.com/IDR/idr.openmicroscopy.org/blob/master/docker.sh


joshmoore:
Thanks, @pwalczysko. Not sure how we go about remembering to _unpin_ this at some point in the future, but makes sense to have it working for now.

PR #644, opened 2023-04-26
Update events April 2023

pwalczysko:
Updating events - 

 - TiM
 - Eastbio
 - NEUBIAS
 

cc @jburel


pwalczysko:
Could we maybe have this one as well please ? Actually, I had the broken npsc link fixed here too :) https://github.com/ome/www.openmicroscopy.org/pull/646#pullrequestreview-1410523196

jburel:
Could you remove the last commit in that case?

pwalczysko:
> Could you remove the last commit in that case?

Done

PR #645, opened 2023-04-26
Bio-Formats 6.13.0 updates

dgault:
Updates to bump Bio-Formats version and add release announcement and new contributors.
Will fail and wont be ready for full review until BF artefacts have been released

PR #646, opened 2023-05-03
Update broken npsc link

dgault:
While running the tests for https://github.com/ome/www.openmicroscopy.org/pull/645 it highlighted that the  http://npsc.ac.uk/ link was broken. The updated link is the closest match I could find for a replacement.


sbesson (approved):

jburel:
I have created a ticket to fix the current redirect in place

PR #665, opened 2023-08-07
ome-model point to latest. 

jburel:
This should redirect to readthedocs stable
i.e. https://ome-model.readthedocs.io/en/stable/


_config.yml
```diff
@@ -49,7 +49,7 @@ omecommoncpp:
   version: 5.5.0
 
 omemodel:
-  version: 6.0.0
+  version: 6.0.1
```
sbesson, line 52: Can this variable safely be dropped if all usages have been replaced internally?

jburel: No we can't since it points to the source under downloads.
The new version have not been synchronised under downloads.
see 
```
"https://downloads.openmicroscopy.org/ome-model/{{ site.omemodel.version }}/source/ome-model-{{ site.omemodel.version }}.zip
```

sbesson: is this the only remaining location? If so, I'd be tempted to hardcode the version in the relevant page (esp. since this product is unmaintained) and remove it from the top-level configuration file

jburel: I forgot to mention it during last BF call.
Agree with suggestion

sbesson (approved):

PR #757, opened 2024-10-23
Updates for Bio-Formats 8.0.0

melissalinkert:
Preparation for tomorrow's planned release. See also https://github.com/ome/bio-formats-documentation/pull/397.


sbesson:
@jburel this PR is now effectively to be reviewed and published by the OME team.
Note the CI build is failing on 2 types of links:
- https://docs.openmicroscopy.org/bio-formats/8.0.0/ and its sub-pages. The documentation has been officially published on readthedocs but as I understand as a result of the migration of the docs.openmicroscopy.org service, the strategy needs to be modified
- https://www.facebook.com/openmicroscopy which is now 400

jburel:
> @jburel this PR is now effectively to be reviewed and published by the OME team. Note the CI build is failing on 2 types of links:
> 
> * https://docs.openmicroscopy.org/bio-formats/8.0.0/ and its sub-pages. The documentation has been officially published on readthedocs but as I understand as a result of the migration of the docs.openmicroscopy.org service, the strategy needs to be modified

What strategy are you refereeing to? 
The ``.htacess`` with the redirect has been in place for now several releases and was used by Apache
We now have a script that will update the nginx config.
Are you thinking of something else?

> * https://www.facebook.com/openmicroscopy which is now 400

I don't think that this is due to the migration. Not something I was aware of

sbesson:
> What strategy are you refereeing to?
The .htacess with the redirect has been in place for now several releases and was used by Apache
We now have a script that will update the nginx config.
Are you thinking of something else?

I have no specifics other than the URL above was not functional immediately after tagging `bio-formats-documentation`. If you fixed the issue with another script updating the NGINX configuration, that's great.

> I don't think that this is due to the migration. Not something I was aware of

There is no relationship with the migration but this failure is holding the review of this PR by the OME team (and its deployment eventually). Should we add it the exclude list in https://github.com/ome/www.openmicroscopy.org/blob/f4b135b6788382640ee1cc59841fe528f6702aa3/build.sh#L11 and let you guys review as a follow-up?

jburel:
It was never available after tagging. The job runs every hour 
```
@hourly (cd /uod/idr/scratch/management_tools/bin && bash release_combine.sh)
```
The new job was not part of cron job since I finished it this morning.
If we want to run more frequently we can
Downloads was updated almost immediately since the tag was pushed before the top of the hour.

jburel:
https://sourceforge.net/p/icu/mailman/message/35232691/ is also failing.I just noticed that it failed on a previous PR (minute)

_posts/2024-10-24-bio-formats-8-0-0.md
```diff
@@ -0,0 +1,72 @@
+---
+layout: post
+title: Release of Bio-Formats 8.0.0
+intro-blurb: The OME team is pleased to announce the release of Bio-Formats 8.0.0
+---
+
+Today we are releasing Bio-Formats 8.0.0 which includes several performance improvements, expansion of the precompressed tile feature, and a new batch processing option in the showinf tool.
```
jburel, line 7:
```suggestion
Today we are releasing Bio-Formats 8.0.0 which includes several performance improvements, expansion of the precompressed tile feature, and a new batch processing option in the `showinf` tool.
```

jburel (approved):

PR #775, opened 2025-01-29
Update jquery to 3.7.1

will-moore:
It looks like we are using `Foundation.js` version `6.3.1`.
https://github.com/ome/www.openmicroscopy.org/blob/f841698e444599d2d49859bbf7c4acd6e01574e4/js/vendor/foundation.js#L5

This has a jQuery dependency of `"jquery": ">=2.2.0",` at https://github.com/foundation/foundation-sites/blob/v6.3.1/package.json

So we should be fine with updating to latest jQuery (3.7.1).

It looks like we were previously loading 2 versions of jQuery. One from https://code.jquery.com/jquery-2.1.4.min.js and `v2.2.4` hosted internally.

I have used the `min` version of jquery: 88 kb instead of uncompressed version at 285 kb.


pwalczysko:
Thank you @will-moore !

This should now be deployed on https://snoopycrimecop.github.io/www.openmicroscopy.org/ - @will-moore do you have a way to check that your changes are indeed on https://snoopycrimecop.github.io/www.openmicroscopy.org/ now please ?

pwalczysko:
@will-moore Edited previous comment ^^^ - the deployment of your changes lgtm. Lets test the result properly tomorrow.

will-moore:
Testing the site deployed above - all seems to be working fine.
But there is a console error on page load which is hard to understand from the `min.js` stack trace:

Screenshot 2025-01-30 at 09 44 53

I'll revert to the un-compressed version and see if it helps to understand.

will-moore:
The error above was coming from an out-dated version of `responsive-tables.js` that was left over from foundation 2.4.1 (which wasn't compatible with the updated jquery). Having removed that (and the corresponding .css) the error has gone. Tested alongside the existing site, comparing pages with tables (we don't have many) and everything looks fine.

jburel:
Thanks @will-moore for investigating

pwalczysko (approved):

pwalczysko:
Thanks @will-moore . Merging.

PR #785, opened 2025-03-25
Update outreach

pwalczysko:
Adding the upcoming outreach event.


pwalczysko:
Could we consider merging this prior to the upcoming workshop ?

PR #821, opened 2025-08-27
Add acknowledgement to Advisories too

pwalczysko:
as discussed with @sbesson  - adding acknowledgement of the reporter also to the Advisory


pwalczysko:
built at https://snoopycrimecop.github.io/www.openmicroscopy.org/security/advisories/2025-54791/

PR #824, opened 2025-09-16
Add advisories for CVEs 2022-22970 and 2022-22971

sbesson:
See https://spring.io/security/cve-2022-22970/ and https://spring.io/security/cve-2022-22971/

These two vulnerabilities have been flagged by security reporting tools running against the latest version of OMERO.server 5.6.15 but OMERO is unaffected as it is does not rely on either of the functionalities.

### DIFF
--- a/security/advisories/2022-22970/index.md
+++ b/security/advisories/2022-22970/index.md
@@ -1,0 +1,17 @@
+---
+layout: security-advisory
+filename: security
+title: CVE-2022-22970 ("Spring Framework DoS via Data Binding to MultipartFile or Servlet Part")
+---
+
+Major news carriers have been reporting on the
+[Spring Framework DoS Vulnerability](https://spring.io/security/cve-2022-22970/)
+in Java applications that utilize Spring.
+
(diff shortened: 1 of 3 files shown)


chris-allan (approved):

will-moore:
Slightly confusing to start with "Major news carriers have been reporting..." which gives the impression this is a recent CVE, but both of these are from 2022. But I understand we're just using standard layout, which is fine.

joshmoore:
I could also see having a different template for when we retroactively check something, but no strong feeling. Thanks for adding, @sbesson! 🙌🏽

sbesson:
Absolutely, I have copied the template originally introduced in https://www.openmicroscopy.org/2021/12/13/log4j-issue.html  but I could see alternative template and retrospective reformatting for these security assessments.

Although I am supportive of the idea, this is not something I intend to add as part of this PR (or even in the near future) though. Given the importance and impact for anyone in the community in charge of deploying OMERO, I'd vote for having the content published as advisory as soon as possible.

PR #868, opened 2026-05-01
Ignore broad

pwalczysko:
Ignore link which is 403

cc @sbesson


sbesson (approved):
Thank you @pwalczysko

PR #867, opened 2026-04-30
Added TissueGnostics GmbH

muhanadz:
Adds TissueGnostics GmbH to the "Organizations supporting OME-TIFF" section of the commercial partners page.

TissueGnostics indicates support for OME-TIFF on the TissueFAXS Imaging Suite page: https://www.tissuegnostics.com/tissuefaxs-platform/tissuefaxs-imaging-suite (under `Open Data Formats` tag), so this change adds them to the existing list with a link to the website.

PR #869, opened 2026-05-01
Fix broken link for OME 2020 speaker

sbesson:
The LinkedIn URL for one of the OME 2020 speakers went 404, replacing it by an institutional URL. There is probably a discussion of whether checking all these historical links is a viable strategy but this is outside the scope of this fix


pwalczysko:
uh, the LinkedIN link fixed here is working for me now (the linkchecker concurs, but fails on https://github.com/ome/omero-iviewer/blob/master/CHANGELOG.md)

sbesson:
Maybe it was intermittent. Closing to focus on the review of #867 and we can always reopen if needed

PR #870, opened 2026-05-05
Fix two links

pwalczysko:
- ignore en.cppreference.com (403)
- Replace the 404-ing link in iviewer section with pointer to releases

cc @sbesson (to fix https://github.com/ome/www.openmicroscopy.org/actions/runs/25383511321)


pwalczysko:
The link on https://snoopycrimecop.github.io/www.openmicroscopy.org/omero/iviewer/ is working as expected. Build is green.

sbesson (approved):

PR #879, opened 2026-07-02
Fix links

pwalczysko:
Fixes 2 links in Events to older workshop events whose sites have been deprecated in the meantime.
These 2 old links were also http.

Replacing with links to the presentations of those 2 workshops on our downloads.op...py.org pages.


pwalczysko:
merging to fix build

PR #881, opened 2026-07-10
Add Glencoe to teams page

erindiel:
Discussed with @joshmoore to add Glencoe and its representatives to the https://www.openmicroscopy.org/teams/ page. Photos were pulled from https://www.glencoesoftware.com/about/team/

@joshmoore can you confirm that this is the correct strategy to extend the page with a Glencoe section?

I can then add bios and assign additional reviewers.


joshmoore:
To the extent I can check json on my phone, it looks great. 😜 Thank you! I'll check more closely ASAP.

pwalczysko:
built at https://snoopycrimecop.github.io/www.openmicroscopy.org/teams/

pwalczysko:
Screenshot 2026-07-13 at 09 59 11


The empty line ^^^ is probably caused by an empty field in the json for "Institution" - maybe we could fix somehow ?

erindiel:
Thanks @pwalczysko ! The problem that I was trying to solve was the duplication of `Glencoe Software` if I included the `name` in the `memberOf` block.

Screenshot 2026-07-13 at 9 38 06 AM

But maybe I can exclude this block altogether? I was originally following the pattern of the other Teams. See https://github.com/ome/www.openmicroscopy.org/pull/881/commits/2ad464dbffe9f45c9be87b17c8cf1b2a085c8fa3 as the potential solution.

Glencoe bios have been added in https://github.com/ome/www.openmicroscopy.org/pull/881/commits/fac844cf881e448091653bd42bb77c5c63775eec, again taken from https://www.glencoesoftware.com/about/team/

FYI @melissalinkert @knabar @chris-allan @emilroz (Seb is OOO, so we will have to act honorably in his absence 😄 )

pwalczysko:
> But maybe I can exclude this block altogether? I was originally following the pattern of the other Teams. See [2ad464d](https://github.com/ome/www.openmicroscopy.org/commit/2ad464dbffe9f45c9be87b17c8cf1b2a085c8fa3) as the potential solution.

@erindiel : Please do not worry too much. As discussed with @joshmoore at a meeting 2hrs ago, we would need to adjust the conditional in the generation of the html so that it skips the empty line when the field is unset. 
Other solutions are a bit ... less elegant. 

I will try to get this passed by @jburel tomorrow morning. lgtm

